PR #2, opened 2020-07-08
support create TCP connection with timeout

vkill:
Hi stjepang

Please review it. thank you.


vkill:
Oh, maybe I can use `futures::select` macro.

stjepang:
Thanks for the PR and sorry for late reply! I don't think timeouts are necessasry for establishing TCP connections because async Rust already provides tools for creating timeouts on arbitrary futures.

And there is now an example on how to use timeouts in the crate root :) https://docs.rs/async-io/0.1.11/async_io/#examples

vkill:
Oh, that's great. thank you for your hard work.

PR #5, opened 2020-07-22
skip optimistic read/write when there already are wakers registered

Keruspe:
Naive attempt at fixing #3

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -518,10 +518,14 @@ impl<T> Async<T> {
     pub async fn read_with<R>(&self, op: impl FnMut(&T) -> io::Result<R>) -> io::Result<R> {
         let mut op = op;
         loop {
-            match op(self.get_ref()) {
-                Err(err) if err.kind() == io::ErrorKind::WouldBlock => {}
-                res => return res,
+            // If there are no blocked readers, attempt the read operation.
+            if !self.source.readers_registered() {
+                match op(self.get_ref()) {
+                    Err(err) if err.kind() == io::ErrorKind::WouldBlock => {}
+                    res => return res,
+                }
             }
+            // Wait until the I/O handle becomes readable.
             optimistic(self.readable()).await?;
         }
     }
@@ -554,10 +558,14 @@ impl<T> Async<T> {
     ) -> io::Result<R> {
         let mut op = op;
         loop {
-            match op(self.get_mut()) {
-                Err(err) if err.kind() == io::ErrorKind::WouldBlock => {}
-                res => return res,
+            // If there are no blocked readers, attempt the read operation.
+            if !self.source.readers_registered() {
+                match op(self.get_mut()) {
+                    Err(err) if err.kind() == io::ErrorKind::WouldBlock => {}
+                    res => return res,
+                }
             }
+            // Wait until the I/O handle becomes readable.
             optimistic(self.readable()).await?;
         }
     }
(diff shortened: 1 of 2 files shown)


Keruspe:
Actually this feels simpler and should do the trick

Keruspe:
So, with rust 1.39 everything works fine, but with stable/beta/nightly most tests timeout because of poll_read/poll_write being called in a loop and never being awoken.
Not sure:
- why they are not being awoken
- why poll_read/poll_write are being called in a loop despite never being awoken
- why the behaviour is different with rust 1.39 and newer ones (I wonder which is the first failing version)

Keruspe:
That was just me misunderstanding stuff.
Works fine now apart from the security check not launching properly?

stjepang:
This seems to be correct but it's a bit unfortunate how every invocation of `{read,write}_with{,_mut}` grabs the lock before doing anything. I'm almost certain this will be noticeable in benchmarks.

I would instead suggest having an `AtomicUsize` inside `Source` with two bits - one indicating that there are wakers interested in reading, and one indicating that there are wakers interested in writing.

Then we could just do an atomic load to see if the `op` should be attempted or if we should go straight to awaiting `readable()` or `writable()`.

Keruspe:
Something like this?

There's probably some bikeshedding needed wrt naming, and I've never really been comfortable with Ordering so I just put SeqCst everywhere as a "safe bet"

Keruspe:
Hmm, after rebasing from 03b5ba6efa4f916eadd075947fcc59186aebc4f4 to d3b5cd401e299c9d73d1c8400c4af9c2335c9b46 it seems to hang now though the logic is the same, will dig into it later

Keruspe:
Fixed.

stjepang:
Thank you! This looks good now, I'm only a bit confused by what it means when `writers_registered()` returns `true`. My first guess was that it means "there are some writers registered", but it seems it actually means "there are no writers"?

Keruspe:
Hmm

We start with `wakers_registered = 0`.
When we enqueue a writer for later notification, we basically do `wakers_registered |= WRITERS_REGISTERED` to set this flag.
When we dequeue all waiting writers we do `wakers_registered &= !WRITERS_REGISTERED` to unset this flag.
When we check if there are some writers registered, we do `wakers_registered & WRITERS_REGISTERED == WRITERS_REGISTERED` which should be true when the flag is set or flase when it isn't? Or am I missing something?

src/reactor.rs
```diff
@@ -427,6 +435,9 @@ struct Wakers {
 }
 
 impl Source {
+    const READERS_REGISTERED: u8 = 1 << 1;
+    const WRITERS_REGISTERED: u8 = 1 << 2;
```
stjepang, line 439:
```suggestion
    const READERS_REGISTERED: u8 = 1 << 0;
    const WRITERS_REGISTERED: u8 = 1 << 1;
```

src/lib.rs
```diff
@@ -516,6 +516,9 @@ impl<T> Async<T> {
     /// # std::io::Result::Ok(()) });
     /// ```
     pub async fn read_with<R>(&self, op: impl FnMut(&T) -> io::Result<R>) -> io::Result<R> {
+        if self.source.readers_registered() {
```
stjepang, line 519: Should this be `if !self.source.readers_registered()`?

So if there are readers waiting on this I/O object, then we can optimistically try a read operation. But if there are readers waiting on it, then we should just skip it and go register ourselves.

Keruspe: The line below is not the optimistic read. If we have registered readers, we wait for the source to be readable (thus registering ourselves) before attempting the read for the first time

stjepang: Ah you're right. Sorry, I think the `optimistic()` threw me off here. 😅

stjepang: Could you perhaps move that check into the loop, so something like:

```rust
// If there are no blocked readers, attempt the read operation.
if !self.source.readers_registered() {
    match op(self.get_ref()) {
        // ...
    }
}
// Wait until the I/O handle becomes readable.
optimistic(self.readable()).await?;
```

Keruspe: Done

PR #7, opened 2020-07-26
Add back Timer::at

Matthias-Fauconneau:
Timer::add is useful to create a stream of repeating events :
https://github.com/Matthias-Fauconneau/ui/blob/master/src/window.rs#L124-L126

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,6 +68,14 @@ pub struct Timer {
 }
 
 impl Timer {
+    /// Creates a timer that expires at the specified instant in time.
+    pub fn at(when: Instant) -> Timer {
+        Timer {
+            id_and_waker: None,
+            when,
+        }
+    }
+
     /// Creates a timer that expires after the given duration of time.
     ///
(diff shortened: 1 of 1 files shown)


stjepang:
How would you feel about renaming this method to `new_at()`?

I'm trying to think of a consistent naming scheme, perhaps this could work:

```rust
impl Timer {
    fn new(dur: Duration) -> Timer;
    fn new_at(when: Instant) -> Timer;

    fn reset(dur: Duration);
    fn reset_at(when: Instant);
}
```

Matthias-Fauconneau:
I don't have any strong feelings about the naming. I liked at/after. I
suppose after got renamed to new because it's the default use.

I can find constructors without new in the name, I think there are common
examples of this pattern IIRC.

I'm not using resets.

stjepang:
I'm going to merge this. I think I also liked `Timer::at()` and `Timer::after()` more that what we have now. Perhaps in the future there could also be `Timer::interval()` and so on.

Issue #6, opened 2020-07-24
Performance issue with `Timer`


Byron:
This 'regression' appears a while ago in `criner`, and apparently went unnoticed when switching from `tokio::time::Delay` to `async_io::Timer`.

<img width="1432" alt="Screenshot 2020-07-24 at 10 17 31" src="https://user-images.githubusercontent.com/63622/88357075-9795f180-cd9c-11ea-8f9e-39d1e63a47c4.png">

The picture above shows a `criner` process doing nothing, consuming 10 to 15% of CPU for displaying 4 progress bars which update once per second. The expensive part of this used to be the drawing itself, but now it appears to be dominated by the condvar in `parking`.

In `smol` 0.1.5 using `smol::Timer`, this looked like that:

<img width="1440" alt="Screenshot 2020-07-24 at 11 00 55" src="https://user-images.githubusercontent.com/63622/88357569-5a326380-cd9e-11ea-9a19-f99a46cbe77d.png">

All that shows up is the drawing of the TUI.

Having had a look myself I am absolutely puzzled as to why that is - the code looks very similar.

**Any hint at how to solve this would be greatly appreciated.**
Thank you

<details>

#### Meta

* ['utils.rs' on a criner branch that works as expected, without performance issues](https://github.com/the-lean-crate/criner/blob/378f7d28bf8591ecaca82cd1f74c382eb9929b9d/criner/src/utils.rs#L41)
   * It's noteworthy that `criner` didn't change in the relevant portions of the code [responsible for the timer invocations](https://github.com/the-lean-crate/criner/compare/without-performance-regression...master#diff-1dcf5476e374956c4bac4ecfde88ea32) _see utils.rs_.

<img width="898" alt="Screenshot 2020-07-24 at 11 20 58" src="https://user-images.githubusercontent.com/63622/88357989-daa59400-cd9f-11ea-9254-354723f64e94.png">

* [the version of 'utils.rs' that suffers from 'slow' timers](https://github.com/the-lean-crate/criner/blob/7c27b064c1d3359daff8ef3cf70c72474211be70/criner/src/utils.rs#L43)

stjepang:
This should now be fixed in v0.1.7 - can you check?

Byron:
Thanks a lot for your help! I think fixing this one will be important so that async-std doesn't get swamped with similar performance related tickets _if_ they use the same timer implementation.

[...]

PR #8, opened 2020-07-30
Fix high CPU usage by the async-io thread

stjepang:
@Byron wanna check if this fixes the issue reported in #6?

The async-io thread is supposed to only "help" push things forward when `Parker`s are not actually parking or when there are no `Parker`s at all. Previously, the async-io thread would be too aggressive, competing with `Parker`s to grab the lock on the `Reactor` too often, rather than giving priority to `Parker`s and only waking up when the reactor is idling.

This could be further improved. On Unix systems, we could use kqueue timers or timerfds to not wake up the async-io thread at all unless the reactor hasn't been polled in X milliseconds. But I hope this PR is good enough for now.

### DIFF
--- a/src/parking.rs
+++ b/src/parking.rs
@@ -489,7 +489,8 @@ impl Reactor {
             // Spawn a helper thread driving the reactor.
             //
             // Note that this thread is not exactly necessary, it's only here to help push things
-            // forward if there are no `Parker`s around.
+            // forward if there are no `Parker`s around or if `Parker`s are just idling and never
+            // parking.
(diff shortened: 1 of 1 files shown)


Byron:
Screenshot 2020-07-31 at 07 21 33

Fantastic - that works! `prodash` is back to the usual CPU usage when idle, which is dominated by the TUI (unfortunately so, but that's for another day and another repository ;)).

stjepang:
Thank you for testing! :)

PR #9, opened 2020-08-07
Factor out cross-platform types into a single module

infinity0:
This will make it a bit easier for me to wrap around the type.

(Specifically for my rate-limiting use case, for the initial PoC I need to create a wrapper that introduces additional buffers for reading and writing (most OSes don't let me access its internal buffers), to know which handles are under load or not. That's not directly related to this PR though.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,14 +60,14 @@ use std::io::{self, IoSlice, IoSliceMut, Read, Write};
 use std::mem::ManuallyDrop;
 use std::net::{Shutdown, SocketAddr, TcpListener, TcpStream, UdpSocket};
 #[cfg(windows)]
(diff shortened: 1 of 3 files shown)


stjepang:
Hey, sorry for late review! I like this change, I would just call this type alias `RawSource` or something like that to avoid confusion with `std::os::windows::io::RawHandle`. :)

infinity0:
@stjepang thanks! updated as your per suggestion & rebased

infinity0:
It turns out, I don't actually need this myself - since I'm not using `polling` I don't need access to the raw handles. However I'll leave this PR open in case you want it anyway.

stjepang:
This is exposing internal trait `AsRawHandle` in the public API, which might make docs a bit confusing. Still, I appreciate the fact that it makes the internal implementation a bit cleaner. I don't know... I might revisit this in the future :)

PR #13, opened 2020-08-20
reactor: properly reset registered wakers state when needed

Keruspe:
If the list of pending wakers grew too big and we woke everything, also drop the bit from wakers_registered


stjepang (approved):
Good catch! Thank you :)

PR #21, opened 2020-08-28
fail fast on `io::ErrorKind::NotConnected`

JayceFayne:
Maybe a comment would be useful? Something like "Workaround to check if there was an error while connecting."?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1222,6 +1222,9 @@ impl Async<UnixStream> {
         // The stream becomes writable when connected.
         stream.writable().await?;
 
+        // On Linux, it appears the socket may become writable even when connecting fails, so we
+        // must do an extra check here and see if the peer address is retrievable.
+        stream.get_ref().peer_addr()?;
         Ok(stream)
     }
 


src/lib.rs
```diff
@@ -1222,6 +1222,7 @@ impl Async<UnixStream> {
         // The stream becomes writable when connected.
         stream.writable().await?;
 
+        stream.get_ref().peer_addr()?;
```
stjepang, line 1225:
```suggestion
        // On Linux, it appears the socket may become writable even when connecting fails, so we
        // must do an extra check here and see if the peer address is retrievable.
        stream.get_ref().peer_addr()?;
```

stjepang (commented):
Looks good! Yeah, let's add a short explanation, suggestion below but feel free to reword it as you wish!

src/lib.rs
JayceFayne: I removed the Linux reference since I think you mentioned the same behaviour on macOS [here](https://github.com/stjepang/async-io/issues/18#issuecomment-681916927) :)

JayceFayne: nvm I just saw [your comment](https://github.com/stjepang/async-io/issues/18#issuecomment-682559099) :smile:

JayceFayne:
Added your suggested comment and rebased!

src/lib.rs
stjepang: Hmmm. It seems that was my mistake and I misspoke in the comment. On macOS, I get `ConnectionRefused` rather than `NotConnected`.

To make sure we're on the same page, this is my understanding:

When `UnixStream` is in non-blocking mode:

* On Linux, `Async::<UnixStream>::connect()` succeeds, but then the first following write errors with `NotConnected`.
* On macOS, `Async::<UnixStream>::connect()` fails immediately with `ConnectionRefused`.

When `UnixStream` is in blocking mode:

* On all systems, `UnixStream::connect()` blocks until the connection is really established.

The workaround here is specifically for Linux. In `Async::<UnixStream>::connect()`, we add an additional check at the end of the function that attempts to retrieve the `peer_addr()` so that we fail immediately rather than on the first following write.

stjepang: When I said "On macOS I get the NotConnected error in all cases." I meant to actually say:

On macOS I get the `ConnectionRefused` error with any of `blocking`, `async-std`, or `async-io`.

stjepang: I don't really know on which OS-es the `peer_addr()` check is necessary, but it probably doesn't hurt having it even on systems that don't need it.

stjepang (approved):
Thank you for all the time and effort spent onto debugging and fixing this issue! :)

src/lib.rs
JayceFayne: True, then let's just leave it like that for now I guess :)

JayceFayne:
Was a pleasure debugging this issue with you! I also want to thank you for creating all the async executor agnostic crates, they helped me in many projects :smile:

stjepang:
Published in v0.2.5

PR #22, opened 2020-08-28
generalise Source to GSource and expose parts of reactor module

infinity0:
This lets me dedupe some code from rs-bwlim - specifically https://github.com/infinity0/rs-bwlim/commit/31483f30ebde343ce2bb828a4c55e46a0fa133df

Pretty sure there should be 0 runtime cost - I used PhantomData & all other changes were just moving functions around.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,7 +79,7 @@ use socket2::{Domain, Protocol, Socket, Type};
 
 use crate::reactor::{Reactor, Source};
 
-mod reactor;
+pub mod reactor;
 
 /// Blocks the current thread on a future, processing I/O events when idle.
 ///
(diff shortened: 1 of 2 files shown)


stjepang:
This comment also applies here: https://github.com/stjepang/async-io/pull/23#issuecomment-688494927

infinity0:
> But also, I feel it probably doesn't make too much sense for async-io and rs-bwlim to share possible. At the end of the day, it's not that much code duplication and those two crates have fairly different goals.

(Replying here because I think this PR is more useful currently, than #23, and has downsides than it.)

It's not a lot of code duplication, but it's code duplication of something that you correctly identified as not being the goal of rs-bwlim. I'm not an expert in how that part of the logic works, and whilst I could indeed become familiar with it after some time, ideally I'd just like to reuse whatever async-io does.

The part that I'm factoring out here and reusing in rs-bwlim is effectively an implementation of a generic reactor, that may or may not be based on epoll. It's not entirely infeasible that someone else would find the logic useful too. For example on a bare-metal system where epoll does not exist, and you have to recreate it, you could use this logic to implement that.

PR #23, opened 2020-08-28
add AnAsync{,Ext} traits to reuse {read,write}_with{,_mut}

infinity0:
This lets me dedupe some more code from rs-bwlim - specifically https://github.com/infinity0/rs-bwlim/commit/163717455b3d9cbd34ea4eb175da11601a486b16

However it is less trivial than #22 - we depend on an unstable nightly-only scary feature of rust - `generic_associated_types`. Also we need `async_trait` which itself a temporary thing, pending much-future advances in rust. So I can totally understand if you don't want it in just yet.

I tried various other strategies but couldn't get them working. The main problem being that due to instance coherence rules, rust does not let you `impl ForeignTrait for T where T: LocalTrait` but this is ideally what we want to do, i.e. `impl AsyncRead for T where T: AnAsyncExt`. So the current PR is the best I came up with.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["asynchronous", "network-programming", "os"]
 readme = "README.md"
 
 [dependencies]
+async-trait = "0.1"
 cfg-if = "0.1.10"
 concurrent-queue = "1.2.0"
 fastrand = "1.3.4"
--- a/examples/linux-inotify.rs
+++ b/examples/linux-inotify.rs
@@ -10,7 +10,7 @@
 fn main() -> std::io::Result<()> {
     use std::ffi::OsString;
 
-    use async_io::Async;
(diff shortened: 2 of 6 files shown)


infinity0:
Oh, another downside is `async_trait` carries a very minor runtime cost since it uses `dyn Future` instead of `impl Future` but this is probably not noticeable.

stjepang:
I don't think this fits into the overall goals of `async-io` - it is now at v1.0 so I'm hesitant of adding new API. And I like keeping things as simple as possible :) 

Also, `async_trait` pulls in `syn` and `quote`, which was entirely eliminated from the dependency tree of `smol` because they make compilation considerably slower.

But also, I feel it probably doesn't make too much sense for `async-io` and `rs-bwlim` to share possible. At the end of the day, it's not that much code duplication and those two crates have fairly different goals.

PR #26, opened 2020-09-09
Less boxing

bugaevc:
Warning: the second commit introduces a slight API break (the returned streams no longer implement `Unpin`). Perhaps you only want to take the first but not the second commit, or perhaps only release the second commit in the next API-breaking release.

### First commit

As far as I can see, the only reason for `Async<T>` to use a `Box<T>` internally was for the `Async<T>` to be `Unpin`, no matter whether `T` itself is `Unpin` or not. But you can achieve that by just explicitly asking for it:

```rust
impl<T> Unpin for Async<T> {}
```

As `Async<T>` never creates a `Pin<&mut T>` or anything like that (that would be useless — most likely `T` is not a `Future` or a `Stream`), it is completely fine to just implement `Unpin` for the aggregate type (`Async<T>`) even if one of the fields is `!Unpin`.

https://doc.rust-lang.org/std/pin/index.html#pinning-is-not-structural-for-field

### Second commit

Again, as far as I can see, the only reason to box the returned streams is for them to be `Unpin`. But in a lot of cases — such as when using stream combinators — it doesn't really matter whether the stream impls `Unpin` or not. It *does* matter when the caller wants to do

```rust
while let Some(item) = stream.next().await {
    ...
}
```
which is, admittedly, a common way to consume a stream in today's Rust (stream combinators are another common way), but it might become less important when we get native `async for` support. In the meantime, if the caller needs for the stream to be pinned, they can do so themselves, with `pin!()`, or by using `Box::pin()` themselves.

But it is a bit less convenient, and kind of breaks the API; that's the trade-off. I still think it's worth it because Rust is All About Avoiding Unnecessary Boxing, but you might disagree.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -853,12 +853,13 @@ impl Async<TcpListener> {
     ///
     /// ```no_run
     /// use async_io::Async;
-    /// use futures_lite::stream::StreamExt;
+    /// use futures_lite::{pin, stream::StreamExt};
     /// use std::net::TcpListener;
     ///
     /// # futures_lite::future::block_on(async {
     /// let listener = Async::<TcpListener>::bind(([127, 0, 0, 1], 8000))?;
-    /// let mut incoming = listener.incoming();
+    /// let incoming = listener.incoming();
+    /// pin!(incoming);
     ///
     /// while let Some(stream) = incoming.next().await {
     ///     let stream = stream?;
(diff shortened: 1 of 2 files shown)


stjepang:
This is all perfectly reasonable! Given that `async-io` v1.0 has only had ~150 downloads on crates.io (most of which are probably mine), we can sneak this minor breaking change in. The chances of this actually breaking anyone's code are low, and compilation errors are easy to fix.

stjepang (approved):
Thanks!

stjepang:
Published v1.0.2

Issue #27, opened 2020-09-11
move `impl Async<std::net::*>` to `async-net`


JayceFayne:
I actually thought about proposing this while investigating #18 and now that `async-io` is at v1.0.x I am not sure if a breaking change directly after v1.0 would be appreciated.

That said I believe it would be a good idea to move all the net related impls meaning `impl Async<TcpListener>`, `impl Async<TcpStream>` and so on to `async-net`. That way we can remove the `socket2` dependency from `async-io` and it would reduce the overall code base. I already looked into this a bit and one obvious downside would be that the current doc tests have to be rewritten since they mostly 12/20 use net related types. Feel free to close this issue if you don't think this is a good idea or if I have simply missed something and it's not possible to implement it that way :smile:

stjepang:
Unfortunately, this is not possible to due to coherence rules: `Async<T>` can only be implemented inside the crate that defined `Async`. Otherwise, we get a compilation error.

Regarding removing `socket2` dependency: I want it removed too, especially considering the fact that `socket2` has a lot of unsafe code and `async-io` is only using a function or two from it. In fact, I want to remove as much unsafe code as possible from the whole dependency tree of `smol`.

JayceFayne:
> Unfortunately, this is not possible to due to coherence rules: `Async<T>` can only be implemented inside the crate that defined `Async`. Otherwise, we get a compilation error.

Ahh I believe you misunderstood me. I know that you cannot `impl Async<T>` in another crate if both `Async` and `T` are types that are not defined by the current crate. What I meant was moving the impls to the already existing impls in `async-net`. E.g. replacing `self.0.accept().await?` [here](https://github.com/stjepang/async-net/blob/master/src/tcp.rs#L135) with the corresponding impl `Async<TcpListener>::accept` [from here](https://github.com/stjepang/async-io/blob/master/src/lib.rs#L843). That way there are no `impl Async<TcpListener>` blocks and the coherence rules don't apply.

> Regarding removing `socket2` dependency: I want it removed too, especially considering the fact that `socket2` has a lot of unsafe code and `async-io` is only using a function or two from it. In fact, I want to remove as much unsafe code as possible from the whole dependency tree of `smol`.

Yes I noticed that too! That's why I would really like to move that code from `async-io` to `async-net`. To be fair, moving the `socket2` dependency from `async-io` to `async-net` would not remove unsafe code from `smol`'s dependency tree...:smile:

stjepang:
[...]

PR #28, opened 2020-09-15
remove socket2 dependency

JayceFayne:
As discussed in #27. This pull request removes the `socket2` and `cfg-if` (was not being used?) dependencies. Most of the code is from rusts standard library and the `socket2` crate. I only implemented features that are actually used in `async-io` atm to keep the code base small.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,29 +13,29 @@ categories = ["asynchronous", "network-programming", "os"]
 readme = "README.md"
 
 [dependencies]
-cfg-if = "0.1.10"
 concurrent-queue = "1.2.2"
 fastrand = "1.3.5"
 futures-lite = "1.4.0"
-libc = "0.2.77"
 log = "0.4.11"
 once_cell = "1.4.1"
 parking = "2.0.0"
 polling = "1.0.1"
-socket2 = { version = "0.3.15", features = ["pair", "unix"] }
 vec-arena = "1.0.0"
 waker-fn = "1.1.0"
 
+[target."cfg(unix)".dependencies]
+libc = "0.2.77"
+
 [target.'cfg(windows)'.dependencies]
 wepoll-sys-stjepang = "1.0.8"
-winapi = { version = "0.3.9", features = ["ioapiset"] }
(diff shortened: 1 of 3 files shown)


stjepang:
That was fast, thank you!! :)

It seems to me we can still remove quite a bit of code (e.g. perhaps some `From<...>` impls) - what do you think? It would be great if we can make the `socket` module almost trivial.

Not that this is a deal-breaker, quite the opposite - I'm cool with merging this now and trimming the code later, but wonder if you have any thoughts :)

JayceFayne:
Yeah, for some reason I just had the urge to implement this :smile: 

> It seems to me we can still remove quite a bit of code (e.g. perhaps some From<...> impls) - what do you think? It would be great if we can make the socket module almost trivial.

Hmmm I really tried to remove unnecessary code. But now that you mention it I think it is possible to remove some code there... :nerd_face:

JayceFayne:
Sorry for spamming the CI. I dont have a windows machine :/

JayceFayne:
I removed 4 impls in the last commit. I think the `From<...>` impls are needed but maybe I am missing some obvious way on how to remove them :smile:

stjepang:
I might push some changes and help in cleaning this up, if you're okay with that? :)

JayceFayne:
sure!

stjepang:
Making progress! This PR now adds only 200 lines of code :)

JayceFayne:
uhhh, I see what you did there :smile: Time to squash commits?

stjepang:
Ok I think I'm done now :) Do you have any other ideas on how to clean this up or are we ready to merge?

JayceFayne:
Alright, I think changing the function order would improve readability

stjepang:
Seems good! Anything else?

JayceFayne:
Should be ready to merge! Thanks for your help on cleaning up and refactoring :heart:

stjepang:
In order to keep this crate clean, I have moved the non-blocking connect code into a new crate:

* https://github.com/stjepang/async-io/commit/784104543bbbbd4d1297884e559bf0f34d2191de
* https://github.com/stjepang/nb-connect

PR #33, opened 2020-10-03
remove `libc` dependency on windows

JayceFayne:
Like [in](https://github.com/stjepang/polling/blob/master/Cargo.toml#L23) `polling` we only need `libc` on unix systems


stjepang (approved):
Thanks!

PR #34, opened 2020-10-03
fix changelog

JayceFayne:
I think this changelog entry is wrong because [the commit message](https://github.com/stjepang/async-io/commit/906de8e99a20a6dc227cb88e7d7aa741d957eb32) that bumped the `polling` version is `Update async-io` :smile:


stjepang (approved):
Indeed :)

PR #38, opened 2020-10-09
poll_readable() and poll_writable()

stjepang:
Supersedes #37

@dvc94ch, I wanted to give suggestions in #37 but this feature turned out to be more complicated than I anticipated. I just went ahead and did all the refactoring that was necessary, even making it so that `AsyncRead` and `AsyncWrite` also depend on `poll_readable()` and `poll_writable()`. Wanna give this PR a try in your project that motivated #36 ?

@taiki-e, I actually decided to take `&self` rather than `&mut self` in those methods with a `# Caveat` section in the documentation that says you shouldn't use them concurrently. There's a similar documented caveat for `AsyncRead` and `AsyncWrite` too. Since `Async` is a low-level primitive, this flexibility makes it easier to build higher-level abstractions like `async-net` on top of it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,7 +76,7 @@ use std::os::windows::io::{AsRawSocket, RawSocket};
 
 use futures_lite::io::{AsyncRead, AsyncWrite};
 use futures_lite::stream::{self, Stream};
-use futures_lite::{future, pin};
+use futures_lite::{future, pin, ready};
 
 use crate::reactor::{Reactor, Source};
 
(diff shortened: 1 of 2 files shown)


dvc94ch:
Works well, thanks!

taiki-e:
> I actually decided to take `&self` rather than `&mut self` in those methods with a `# Caveat` section in the documentation that says you shouldn't use them concurrently. There's a similar documented caveat for `AsyncRead` and `AsyncWrite` too. Since `Async` is a low-level primitive, this flexibility makes it easier to build higher-level abstractions like `async-net` on top of it.

If it's documented, I think it's okay.

By the way, does this caveat mean that the "any number of concurrent I/O operations" mentioned in https://github.com/rust-lang/futures-rs/pull/2139 is no longer supported with `AsyncRead`/`AsyncWrite`?

stjepang:
@taiki-e Yes, that is correct.

dvc94ch:
What's the status of this pr?

stjepang:
Published in v1.2.0

Issue #35, opened 2020-10-05
Interval API


stjepang:
It would be great if `Timer` was not just a `Future` but also a `Stream` so that it could act as a "ticker" delivering events at regular intervals.

Proposed API:

```rust
impl Timer {
    fn interval(period: Duration) -> Timer;
    fn interval_at(start: Instant, period: Duration) -> Timer;

    fn set_interval(&mut self, period: Duration);
    fn set_interval_at(&mut self, start: Instant, period: Duration);
}

impl Stream for Timer {
    type Item = Instant;
}
```

Thoughts or suggestions?

stjepang:
Published v1.3.0

PR #41, opened 2020-11-20
Add interval api.

dvc94ch:
Based on #35, needed to use async-io in more places in libp2p.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -16,14 +16,14 @@ jobs:
         rust: [nightly, beta, stable, 1.39.0]
     steps:
       - uses: actions/checkout@v2
-      
+
       - name: Set current week of the year in environnement
         if: startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macOS')
-        run: echo "::set-env name=CURRENT_WEEK::$(date +%V)"
-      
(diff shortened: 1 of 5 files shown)


dvc94ch:
Looks like ci is having issues

dvc94ch:
@stjepang any comments? It's currently blocking a PR of mine.

stjepang (changes requested):
Thanks for implementing this! The PR looks good, I would just request the following changes:

1. Implement `poll()` by calling `poll_next()`. Treat a regular timer (`at()`/`after()`) as a stream (`interval_at()`) with a period of infinity.
2. Add a unit test to tests/timer.rs

dvc94ch:
Not sure how to construct an infinite `Duration`. Is 100y infinite enough?

stjepang:
Why is the first tick of a timer constructed by `interval()` immediate? `setInterval()` in javscript and `Ticker` in Go don't fire immediately, so I'm inclined to follow the same behavior here...

dvc94ch:
I guess that is what the start is for no? When periodically performing a process, usually you want to do it now too. Also this matches the behaviour of `wasm-timer`.

dvc94ch:
I think I agree, you can easily get the first tick with `Timer::interval_at(Instant::now(), period)`

stjepang (approved):
Thank you!

dvc94ch:
Can you release 1.2.1? Then I can update my PR's to libp2p.

PR #44, opened 2020-12-10
impl AsRef<T> and AsMut<T> for Async<T>

zeenix:
~~These enable generic implementations for `Borrow<T>` and `BorrowMut<T>` to work on `Async<T>`.~~

These enable generic implementations for `AsRef<T>` and `AsMut<T>` to work on `Async<T>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -915,6 +915,18 @@ impl<T> Async<T> {
     }
 }
 
+impl<T> AsRef<T> for Async<T> {
+    fn as_ref(&self) -> &T {
+        self.get_ref()
+    }
+}
+
+impl<T> AsMut<T> for Async<T> {
+    fn as_mut(&mut self) -> &mut T {
+        self.get_mut()
+    }
+}
+
 impl<T> Drop for Async<T> {
     fn drop(&mut self) {
         if self.io.is_some() {


zeenix:
@stjepang or should this be `AsRef` and `AsMut` instead?

zeenix:
The failing CI is not my fault FWIW. :) I think maybe it just needs to be rerun?

zeenix:
> @stjepang or should this be `AsRef` and `AsMut` instead?

@zserik @taiki-e Hi, Do you have any preference on this?

taiki-e:
I think AsRef is better suited for this, but I'm not sure if it's a good idea to implement the trait in the first place.

https://doc.rust-lang.org/nightly/std/convert/trait.AsRef.html

> AsRef has the same signature as Borrow, but Borrow is different in few aspects:
> 
> - Unlike AsRef, Borrow has a blanket impl for any T, and can be used to accept either a reference or a value.
> - Borrow also requires that Hash, Eq and Ord for borrowed value are equivalent to those of the owned value. For this reason, if you want to borrow only a single field of a struct you can implement AsRef, but not Borrow.
>
> Note: This trait must not fail. If the conversion can fail, use a dedicated method which returns an Option<T> or a Result<T, E>.


https://doc.rust-lang.org/1.30.0/book/first-edition/borrow-and-asref.html

> Choose Borrow when you want to abstract over different kinds of borrowing, or when you’re building a data structure that treats owned and borrowed values in equivalent ways, such as hashing and comparison.
>
> Choose AsRef when you want to convert something to a reference directly, and you’re writing generic code.

taiki-e:
> The failing CI is not my fault FWIW. :) I think maybe it just needs to be rerun?

FYI: CI failure should be fixed in #46

zeenix:
@taiki-e Thanks so much for the info and also for fixing the CI.

> but I'm not sure if it's a good idea to implement the trait in the first place.

How so? I have a [use-case](https://gitlab.freedesktop.org/zeenix/zbus/-/blob/master/zbus/src/raw/socket.rs#L61) where this would allow me to implement a trait in a generic way once, rather than twice. I think others might have a similar case too.

zeenix:
> > The failing CI is not my fault FWIW. :) I think maybe it just needs to be rerun?
> 
> FYI: CI failure should be fixed in #46

Seems [it didn't](https://github.com/smol-rs/async-io/pull/44/checks?check_run_id=1605149027). :( Probably unrelated to your fix?

taiki-e:
> Seems [it didn't](https://github.com/smol-rs/async-io/pull/44/checks?check_run_id=1605149027). :( Probably unrelated to your fix?

You have to rebase (or merge master). The fix is in master, but not in your branch.

> How so? I have a [use-case](https://gitlab.freedesktop.org/zeenix/zbus/-/blob/master/zbus/src/raw/socket.rs#L61) where this would allow me to implement a trait in a generic way once, rather than twice. I think others might have a similar case too.

I'm not familiar with the current implementation of `Async`, so it might be wrong, but `Async` has additional states, and that may cause problems if working directly with the underlying type. It may be okay if it is explicit (via `get_(ref|mut)`), but implementing AsRef/Borrow allows the user to access the underlying type implicitly.
Also, I haven't read the actual code that you linked, but if it has a method that does a blocking operation, it may not be correct to use the same implementation for both async and sync.

In any case, this allows easy to use sync and async types as the same type. This may increase misuse.

zeenix:
> > Seems [it didn't](https://github.com/smol-rs/async-io/pull/44/checks?check_run_id=1605149027). :( Probably unrelated to your fix?
> 
> You have to rebase (or merge master). The fix is in master, but not in your branch.

Ah, yes lol. Didn't think of that. :-) 

> I'm not familiar with the current implementation of `Async`, so it might be wrong, but `Async` has additional states, and that may cause problems if working directly with the underlying type.

From what I can tell, `Async` is just a wrapper type and it gives you access to the underlying type (as you pointed out) already. I don't foresee any issues from allowing also generic access to the underlying type causing additional problems.

> Also, I haven't read the actual code that you linked, but if it has a method that does a blocking operation, it may not be correct to use the same implementation for both async and sync.

As `Async` puts the underlying socket into non-blocking mode, blocking operations are not possible unless one explicitly puts the socket into blocking mode (which they already can). The code I pointed to, can handle both blocking and non-blocking operations and in case of `Async` it just will always be non-blocking.

taiki-e:
> From what I can tell, `Async` is just a wrapper type and it gives you access to the underlying type (as you pointed out) already. I don't foresee any issues from allowing also generic access to the underlying type causing additional problems.

Yeah, but my point is not whether it is possible, but whether it is preferable to be able to do it implicitly.

> As `Async` puts the underlying socket into non-blocking mode, blocking operations are not possible unless one explicitly puts the socket into blocking mode (which they already can). The code I pointed to, can handle both blocking and non-blocking operations and in case of `Async` it just will always be non-blocking.

Ah, you're right. `Async` can also be used with types that cannot be in non-blocking mode, but I think this is okay because it is not recommended to use it with types that cannot be in non-blocking mode.

zeenix:
> > From what I can tell, `Async` is just a wrapper type and it gives you access to the underlying type (as you pointed out) already. I don't foresee any issues from allowing also generic access to the underlying type causing additional problems.
> 
> Yeah, but my point is not whether it is possible, but whether it is preferable to be able to do it implicitly.

I think what I'm fail to understand is why the `AsRef`  in your opinion, makes it implicit. You still have to call `as_ref()` method to get to the underlying type.

taiki-e:
> I think what I'm failing to understand is why the `AsRef` in your opinion, makes it implicit. You still have to call `as_ref()` method to get to the underlying type.

I was thinking of cases that use `impl AsRef` in arguments. (It is implicit, at least on the user side.)

zeenix:
> I was thinking of cases that use `impl AsRef` in arguments. (It is implicit, at least on the user side.)

Ah, I think I get what you mean but I don't agree. Since `Borrow<T>` is implemented for `T`, I'd see your point for that more but `AsRef<T>` isn't implemented for `T` implicitly so it's all very explicit even in the generic client-side code. In fact now that I've changed to `AsRef`, I can't use it for my original use case in zbus since `AsRef<UnixStream>` isn't implemented for `UnixStream`.

zeenix:
@taiki-e anyway, let me know if you feel strongly about this still and we can get a third opinion from one of the maintainers.

taiki-e:
> but `AsRef<T>` isn't implemented for `T` implicitly so it's all very explicit even in the generic client-side code

> I can't use it for my original use case in zbus since `AsRef<UnixStream>` isn't implemented for `UnixStream`.

Oh, sorry, I misunderstood that even if changed to AsRef, it could be used in the use case you mentioned.

src/lib.rs
```diff
@@ -915,6 +915,18 @@ impl<T> Async<T> {
     }
 }
 
+impl<T> std::convert::AsRef<T> for Async<T> {
```
taiki-e, line 918:
```suggestion
impl<T> AsRef<T> for Async<T> {
```

nit: AsRef/AsMut is included in prelude,

src/lib.rs
```diff
@@ -915,6 +915,18 @@ impl<T> Async<T> {
...
+    fn as_ref(&self) -> &T {
+        self.get_ref()
+    }
+}
+
+impl<T> std::convert::AsMut<T> for Async<T> {
```
taiki-e, line 924:
```suggestion
impl<T> AsMut<T> for Async<T> {
```

taiki-e (approved):
My remaining concerns are my misunderstandings and seem unlikely to happen in practice. So I think this is okay.

zeenix:
Squashing..

PR #52, opened 2021-02-21
Statically guarantee I/O handle will be present

Plecra:
Since the io handle is always expected to exist, holding it in an `Option` was a little confusing. The change requires a line of `unsafe` in the new implementation of `into_inner`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -491,7 +491,7 @@ pub struct Async<T> {
     source: Arc<Source>,
 
     /// The inner I/O handle.
-    io: Option<T>,
+    io: T,
 }
 
 impl<T> Unpin for Async<T> {}
@@ -597,7 +597,7 @@ impl<T: AsRawSocket> Async<T> {
 
         Ok(Async {
             source: Reactor::get().insert_io(sock)?,
-            io: Some(io),
+            io,
         })
     }
 }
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks for the PR! Writing unsafe code related to drops correctly is very difficult and we can easily write unsound code. (The amount of unsafe code can easily increase, even if it's only one line now.) So I would like to stick to using safe code unless this shows a concrete performance improvement.

src/lib.rs
```diff
@@ -663,9 +663,9 @@ impl<T> Async<T> {
     /// # std::io::Result::Ok(()) });
     /// ```
     pub fn into_inner(mut self) -> io::Result<T> {
-        let io = self.io.take().unwrap();
         Reactor::get().remove_io(&self.source)?;
-        Ok(io)
+        let mut this = core::mem::ManuallyDrop::new(self);
```
taiki-e, line 667: This leaks `self.source`, right?

Plecra:
I wouldn't expect this to significantly improve performance 😄 I only think the `Option` type is a bit misleading here, by making it unclear when it can be assumed to be `Some`. A comment on the field should be fine, and maybe using a `MayDrop` alias of `Option`.

PR #64, opened 2021-05-22
Add Readable and Writable futures

taiki-e:


yoshuawuyts (approved):
Neat!

Kestrer:
What happens if the `Readable`/`Writable` futures outlive the `Async`? Could that cause problems?

taiki-e:
Hmm...
```text
Error: Os { code: 9, kind: Other, message: "Bad file descriptor" }
```

taiki-e:
In this case, the error is expected behavior, but I think I should have left a reference to `self` to prevent users from writing such code...
async-net's TcpStream and UnixStream wraps the Async in Arc, so it would have made more sense to add a method like readable_owned that takes `self: &Arc<Self>` instead...

1.5.0 has only been released about 2 hours ago, so I'll yank 1.5.0...

@Kestrer Thanks for catching this problem.

Kestrer:
Oh no, is there no way to support this? Would it be possible to for example perform a check if the resource has been unregistered and have Readable error out?

taiki-e:
@Kestrer
> Would it be possible to for example perform a check if the resource has been unregistered and have Readable error out?

I think it is possible, but I would like to provide an API that prevents such error if possible. 

In fact, the error that occurs, in that case, can be easily reproduced by more low-level libraries such as polling, so this PR's API is not necessarily bad. So, we could simply add documentation that says "If writable/readable polled after Async dropped, it will result in an error", but I tend to think it is better if this could be prevented at compile time...

(I'm still not convinced it's worth the breaking change, but we found the problem fast enough that I didn't have to worry about the impact of the breaking change, so I yanked it anyway. Sorry if anyone else is affected by this yanking!)

taiki-e:
And here is the API & implementation I'm thinking of: https://github.com/smol-rs/async-io/compare/master...taiki-e/owned 
(TR; DR: change `Readable` and `Writable` to contain a reference to `Async`, and add `Async::{readable_owned, writable_owned}` that take `self: &Arc<Self>`)

I'm not sure about other use cases where this feature was needed, but it is enough for the async-net use case (https://github.com/smol-rs/async-net/compare/master...taiki-e/owned), 

@Kestrer @smol-rs/admins Any thoughts on this?

Kestrer:
I think this is a good idea. `{readable, writable}_owned` should probably take `Arc<Self>` to reduce unnecessary cloning. And it might be cleaner to implement it using a generic `AsRef<Source>`, implemented by both `Arc<Async>` and `&Async`.

taiki-e:
Filed #66. Thanks @Kestrer for the feedback!

PR #66, opened 2021-05-23
Add Async::{readable_owned, writable_owned}

taiki-e:
This replaces yanked #64.

- change `Readable` and `Writable` futures to contain a reference to `Async`
- add `Async::{readable_owned, writable_owned}` that take `self: Arc<Self>`

See https://github.com/smol-rs/async-io/pull/64#issuecomment-846432757 (and the comments before and after) for more.

r? @Kestrer @yoshuawuyts  @smol-rs/admins

### DIFF
--- a/src/reactor.rs
+++ b/src/reactor.rs
@@ -1,6 +1,9 @@
+use std::borrow::Borrow;
 use std::collections::BTreeMap;
+use std::fmt;
 use std::future::Future;
 use std::io;
+use std::marker::PhantomData;
 use std::mem;
 #[cfg(unix)]
 use std::os::unix::io::RawFd;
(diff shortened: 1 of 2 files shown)


src/reactor.rs
```diff
@@ -465,57 +478,109 @@ impl Source {
...
+    }
+}
+
+impl<T> fmt::Debug for OwnedWritable<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("OwnedWritable").finish()
+    }
+}
+
+struct Ready<S: Borrow<crate::Async<T>>, T> {
```
taiki-e, line 560: Originally I wanted to use `AsRef<Source>`, as @Kestrer suggested in https://github.com/smol-rs/async-io/pull/64#issuecomment-846433555, but I could not implement it because we [had already implemented `AsRef<T>` for `Async<T>`](https://github.com/smol-rs/async-io/blob/e82546dbd93f1d44798de934d8c82f8dbd92c42f/src/lib.rs#L920).

src/reactor.rs
```diff
@@ -465,57 +478,109 @@ impl Source {
...
+
+impl<T> fmt::Debug for Readable<'_, T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Readable").finish()
+    }
+}
+
+/// Future for [`Async::readable_owned`](crate::Async::readable_owned).
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct OwnedReadable<T>(Ready<Arc<crate::Async<T>>, T>);
```
Kestrer, line 502: I find it a bit confusing it's `readable_owned` but `OwnedReadable`. Could this be changed to `ReadableOwned`?

taiki-e: Renamed to ReadableOwned/WritableOwned.

PR #69, opened 2021-07-25
Rewrite driver to use a state machine

### LINKED ISSUE #40: Avoid spawning the async-io thread if possible
In most programs that use `async-io`, the user will simply run `block_on` at the top level. However, this will still spawn the `async-io` thread despite it not being necessary. Would it be possible to avoid that?

danieldg:
This passes the responsibility for calling the Reactor between threads
instead of relying on the async-io thread to pick up the abandoned
Reactor.  Threads that are waiting in block_on are likely to acquire the
reactor on their second wait, and the main thread is only woken if there
are no block_on threads polling.

This also adds support for disabling the async-io thread, which allows
applications that only process I/O from within block_on to avoid the
needless thread creation.  It also supports designating the async-io
thread as the sole owner of the Reactor to avoid this pass-the-reactor
behavior, if that is undesired.

fixes #40

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ concurrent-queue = "1.2.2"
 futures-lite = "1.11.0"
 log = "0.4.11"
 once_cell = "1.4.1"
-parking = "2.0.0"
 polling = "2.0.0"
 slab = "0.4.2"
 socket2 = { version = "0.4.0", features = ["all"] }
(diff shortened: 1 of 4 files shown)


taiki-e:
Closing per decision in https://github.com/smol-rs/async-io/issues/40.

Thanks anyway for the PR!

PR #76, opened 2022-01-26
Use rustix instead of direct calls to libc.

sunfishcode:
Use the [rustix] syscall wrapper crate to factor out error handling and
unsafe system calls. This reduces the amount of unsafe code here, and is
a step towards factoring it out entirely once [`AsFd`] is stabilized and
can replace `AsRawFd` for these kinds of uses.

This does require incrementing the minimum required Rust version to 1.48.
Please feel free to decline this PR if you don't wish to take on these new
requirements.

[rustix]: https://crates.io/crates/rustix/
[`AsFd`]: https://doc.rust-lang.org/stable/std/os/unix/io/trait.AsFd.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ name = "async-io"
 version = "1.12.0"
 authors = ["Stjepan Glavina <stjepang@gmail.com>"]
 edition = "2018"
-rust-version = "1.46"
+rust-version = "1.48"
 description = "Async I/O and timers"
 license = "Apache-2.0 OR MIT"
 repository = "https://github.com/smol-rs/async-io"
@@ -33,7 +33,7 @@ waker-fn = "1.1.0"
 autocfg = "1"
 
 [target."cfg(unix)".dependencies]
-libc = "0.2.77"
+rustix = { version = "0.36.0", features = ["fs"] }
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.42", features = ["Win32_Networking_WinSock"] }
@@ -49,7 +49,6 @@ tempfile = "3"
 
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 inotify = { version = "0.10", default-features = false }
-nix = { version = "0.25", default-features = false }
 timerfd = "1"
 
 [target.'cfg(windows)'.dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ waker-fn = "1.1.0"
 autocfg = "1"
 
 [target."cfg(unix)".dependencies]
-rustix = { version = "0.36.0", features = ["fs"] }
+rustix = { version = "0.36.0", default-features = false, features = ["std", "fs"] }
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.42", features = ["Win32_Networking_WinSock"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,18 +25,13 @@ futures-lite = "1.11.0"
 log = "0.4.11"
 parking = "2.0.0"
 polling = "2.0.0"
+rustix = { version = "0.36.0", default-features = false, features = ["std", "fs"] }
 slab = "0.4.2"
 socket2 = { version = "0.4.2", features = ["all"] }
 waker-fn = "1.1.0"
 
 [build-dependencies]
 autocfg = "1"
-
-[target."cfg(unix)".dependencies]
-rustix = { version = "0.36.0", default-features = false, features = ["std", "fs"] }
-
-[target.'cfg(windows)'.dependencies]
-windows-sys = { version = "0.42", features = ["Win32_Networking_WinSock"] }
 
 [dev-dependencies]
 async-channel = "1"
--- a/examples/linux-timerfd.rs
+++ b/examples/linux-timerfd.rs
@@ -14,6 +14,7 @@ fn main() -> std::io::Result<()> {
 
     use async_io::Async;
     use futures_lite::future;
+    use rustix::fd::BorrowedFd;
     use timerfd::{SetTimeFlags, TimerFd, TimerState};
 
     /// Sleeps using an OS timer.
@@ -24,7 +25,12 @@ fn main() -> std::io::Result<()> {
 
         // When the OS timer fires, a 64-bit integer can be read from it.
         Async::new(timer)?
-            .read_with(|t| nix::unistd::read(t.as_raw_fd(), &mut [0u8; 8]).map_err(io::Error::from))
+            .read_with(|t| {
+                // Safety: Assume `as_raw_fd()` returns a valid fd; when `AsFd`
+                // is stabilized, we can remove this unsafe and simplify.
+                let fd = unsafe { BorrowedFd::borrow_raw(t.as_raw_fd()) };
+                rustix::io::read(fd, &mut [0u8; 8]).map_err(io::Error::from)
+            })
             .await?;
         Ok(())
     }
--- a/examples/linux-timerfd.rs
+++ b/examples/linux-timerfd.rs
@@ -26,8 +26,10 @@ fn main() -> std::io::Result<()> {
         // When the OS timer fires, a 64-bit integer can be read from it.
         Async::new(timer)?
             .read_with(|t| {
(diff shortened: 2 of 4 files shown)


notgull:
I support this in theory; `async-io` is only `unsafe` due to these lines of code. However, it looks like `rustix` is broken on Android. I'd support using `nix`, or even `socket2` to make this cross platform.

sunfishcode:
Ah, this PR was using rustix 0.35.6-beta.2. The Android build error was fixed in the 0.35.6 release. I've now updated the PR.

notgull:
Are there any blockers to this being merged? I see how the extra dependency might give pause, but it's probably worth purging the remaining unsafe code in this crate, at least on Unix.

examples/linux-timerfd.rs
```diff
@@ -24,7 +25,12 @@ fn main() -> std::io::Result<()> {
 
         // When the OS timer fires, a 64-bit integer can be read from it.
         Async::new(timer)?
-            .read_with(|t| nix::unistd::read(t.as_raw_fd(), &mut [0u8; 8]).map_err(io::Error::from))
+            .read_with(|t| {
+                // Safety: Assume `as_raw_fd()` returns a valid fd; when `AsFd`
+                // is stabilized, we can remove this unsafe and simplify.
+                let fd = unsafe { BorrowedFd::borrow_raw(t.as_raw_fd()) };
```
notgull, line 31: `AsFd` is stabilized now, so I think we can eliminate this `unsafe`.

notgull (approved):
Truth be told, this looks good to me.

With `rustix` in the dependency tree, it may be a good idea to also reimplement some other code in terms of `rustix`. For instance, the package has support for safe [`epoll`](https://docs.rs/rustix/latest/rustix/io/epoll/index.html), which means we may be able to reimplement our `epoll` backend in `polling` in terms of it to reduce the amount of unsafe code we use.

There are three main points of contention, from what I see:

- Is there a reason to use this instead of [`nix`](https://crates.io/crates/nix)? While `rustix` is more popular, `nix` may also appear in the dependency tree.
- `rustix` adds the `bitflags` dependency to `smol`'s dependency tree, which compiles fairly quickly but is still another brick in the wall.
- Is the unsafe code we eliminate worth it in the end, since it's just replicated in `rustix` anyways?

@smol-rs/admins This may be worth a more thorough discussion to determine what direction we'd like to move in.

taiki-e:
@sunfishcode 

Thanks for the PR! I'm basically agreed with merging this, but have a few questions.

- Are outline asm implementations for x86/arm/riscv tested? I could not find pre-1.59 tests in the `rustix`'s CI files.
  (FWIW, if they are not tested, I think you can test those implementations with the latest rustc by inverting the `asm` cfg, [like tokio does](https://github.com/tokio-rs/tokio/blob/28ec4a6161210a1ffb66cdcd7958b86b4b78b54e/tokio/build.rs#L97-L100).)
- Looking at the inline asm implementation of `rustix`, it seems that `(in)lateout` is used even where it seems unnecessary. At first glance, it looks fine, but you might want to make sure it is not affected by a bug like https://github.com/rust-lang/rust/issues/101346.
- If we merge this, as notgull mentioned we also want to replace the direct dependencies on `libc` in other smol-rs crates ([`polling`](https://github.com/smol-rs/polling) and [`async-process`](https://github.com/smol-rs/async-process), IIRC) with `rustix`. I don't see a problem with `async-process`, but `polling` uses some platform-specific APIs such as bsd's kqueue, solarish's event ports, so probably can't switch to `rustix` right away except on linux.
  Is there any plan to support those non-linux platform-specific APIs in `rustix`? Or is patch to add support for them acceptable?

---

@notgull 

> * Is there a reason to use this instead of [`nix`](https://crates.io/crates/nix)? While `rustix` is more popular, `nix` may also appear in the dependency tree.

Given nix's MSRV policy ([current MSRV is 1.56](https://github.com/nix-rust/nix#minimum-supported-rust-version-msrv)), it is difficult to have `nix` as a dependency of smol-rs crates.

> * Is the unsafe code we eliminate worth it in the end, since it's just replicated in `rustix` anyways?

Personally, I think the optimization such as uses of raw system calls to be the important benefit of this PR, not only the removal of unsafe code. (If the removal of unsafe code were the only benefit of this PR, I might not have agreed with this PR.)

Cargo.toml
```diff
@@ -33,7 +33,7 @@ waker-fn = "1.1.0"
 autocfg = "1"
 
 [target."cfg(unix)".dependencies]
-libc = "0.2.77"
+rustix = { version = "0.36.0", features = ["fs"] }
```
taiki-e, line 36: Is it possible to disable the `use-libc-auxv` feature?

sunfishcode: Yes; I've now added a patch that does that.

src/lib.rs
```diff
@@ -598,21 +598,19 @@ impl<T: AsRawFd> Async<T> {
...
-            }
-            if res == -1 {
-                return Err(io::Error::last_os_error());
-            }
-        }
+        //
+        // Safety: We assume `as_raw_fd()` returns a valid fd. When
+        // `AsFd` is stabilized and `TimerFd` implements it, we can
+        // remove this unsafe and simplify this.
+        let fd = unsafe { rustix::fd::BorrowedFd::borrow_raw(raw) };
```
notgull, line 608: The Windows backend uses the `FIONBIO` syscall to move the socket into non-blocking mode. It seems that `rustix` also exposes `FIONBIO` for Windows, could we use that to eliminate that unsafe block as well?

sunfishcode:
>     * Are there tests for [_outline_](https://github.com/bytecodealliance/rustix/tree/main/src/backend/linux_raw/arch/outline) asm implementations for x86/arm/riscv? I could not find pre-1.59 tests in the `rustix`'s CI files.
>       (FWIW, if they are not tested, I think you can test those implementations with the latest rustc by inverting the `asm` cfg, [like tokio does](https://github.com/tokio-rs/tokio/blob/28ec4a6161210a1ffb66cdcd7958b86b4b78b54e/tokio/build.rs#L97-L100).)

Ah, good catch. There are `cargo check` tests for 1.48, but not `cargo test`.  And it looks like rustix has acquired some tests which are not compatible with Rust 1.48, so it's not trivial to just enable it. I'll look into fixing that. 

Right now, I can confirm that with those tests commented out, the rest of the tests do still pass on 1.48 on all archtectures that have outline asm implementations.

>     * Looking at the inline asm implementation of `rustix`, it seems that `(in)lateout` is used even where it seems unnecessary. At first glance, it looks fine, but you might want to make sure it is not affected by a bug like [Incorrect handling of lateout pairs in inline asm rust-lang/rust#101346](https://github.com/rust-lang/rust/issues/101346).

Thanks for the heads-up here. Rustix doesn't have that specific pattern, with `lateout` with a virtual register, and has never seen a bug of this sort in any Rust version. But I'll also investigate this more.

>     * If we merge this, as notgull mentioned we also want to replace the direct dependencies on `libc` in other smol-rs crates ([`polling`](https://github.com/smol-rs/polling) and [`async-process`](https://github.com/smol-rs/async-process), IIRC) with `rustix`. I don't see a problem with `async-process`, but `polling` uses some platform-specific APIs such as bsd's kqueue, solarish's event ports, so probably can't switch to `rustix` right away except on linux.
>       Is there any plan to support those non-linux platform-specific APIs in `rustix`? Or is patch to add support for them acceptable?

Rustix does have various non-Linux platform-specific APIs, for example [fcopyfile](https://github.com/bytecodealliance/rustix/blob/main/src/fs/fcopyfile.rs#L8). There's no specific plan; things get added as needed. Patches to add more are welcome. Filing issues is good too; polling APIs can be subtle, and I'd be happy to help figure things out.

src/lib.rs
sunfishcode: I've now added a patch which does this.

examples/linux-timerfd.rs
sunfishcode: Almost, but async-io still has an MSRV of 1.48, and I/O safety is stabilized in Rust 1.63. I've now added more to these comments to mention that.

notgull: We don't run the tests on the MSRV build, so it's alright.

taiki-e (approved):
LGTM

PR #77, opened 2022-03-31
add support for uds_windows (Unix domain sockets for Windows 10 build…

moh-eulith:
… 1809, release in 2018, and above)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ libc = "0.2.77"
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.9", features = ["winsock2"] }
+uds_windows = "1.0.1"
 
 [dev-dependencies]
 async-channel = "1.4.2"
@@ -45,5 +46,3 @@ inotify = { version = "0.10", default-features = false }
 nix = "0.23"
 timerfd = "1.1.1"
 
-[target.'cfg(windows)'.dev-dependencies]
-uds_windows = "1"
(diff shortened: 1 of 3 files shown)


moh-eulith:
Fixed formatting.

notgull:
In my opinion, dealing with socket types not in the standard library is out of the scope for `async-io`.

moh-eulith:
@notgull I respectfully disagree. I wasn't trying to set a new precedent here. Just look at the existing Cargo.toml:

```
[target."cfg(unix)".dependencies]
libc = "0.2.77"

[target.'cfg(windows)'.dependencies]
winapi = { version = "0.3.9", features = ["winsock2"] }
```

Clearly, this stuff is highly OS specific, precisely because the abstraction in the Rust standard lib is rather poor. Async-io fills in that void, and is much needed and appreciated for that.

I think the result of adding this speaks for itself: tests/async.rc removes 5 lines of `#[cfg(unix)]` and therefore passes 5 tests without being tied to a particular OS.

taiki-e:
Thanks for the PR! I would prefer not to do this in async-io, as we do not want to include third-party crates in the public API whenever possible.

However, if we can provide the same API as the Unix one, it might be ok to support this as an optional feature of async-net.

moh-eulith:
I was trying to implement this here so it gets inherited by async-std, which is used by another project in a non-portable way.

I'll try to put all the code in async-std and go from there.

PR #80, opened 2022-05-18
The scope of the unsafe block can be appropriately reduced

peamaeq:
Shrink unsafe block for the  Async impl


taiki-e:
Thanks! Could you resolve the format issue by running `cargo fmt`?

taiki-e (approved):

PR #84, opened 2022-08-11
Implement I/O-safe traits

notgull:
This PR implements `AsFd` and `AsSocket` for `Async<T>`. In addition, `TryFrom<OwnedFd/OwnedSocket` is implemented for `Async<T>` and `TryFrom<Async<T>>` is implemented for `OwnedFd/OwnedSocket`, in order to mirror the `From/Into<OwnedFd/OwnedSocket>` impls in `std` but also considering that the operations are fallible in `async-io`.

See also: sunfishcode/io-lifetimes#38

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,14 +25,14 @@ slab = "0.4.2"
 socket2 = { version = "0.4.2", features = ["all"] }
 waker-fn = "1.1.0"
 
+[build-dependencies]
+autocfg = "1"
+
 [target."cfg(unix)".dependencies]
 libc = "0.2.77"
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.9", features = ["winsock2"] }
-
-[features]
-io_safety = []
 
 [dev-dependencies]
 async-channel = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,9 @@ libc = "0.2.77"
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.9", features = ["winsock2"] }
 
+[features]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -31,6 +31,9 @@ libc = "0.2.77"
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.9", features = ["winsock2"] }
 
+[features]
+io_safety = []
```
taiki-e, line 35: A comment explaining what this feature does would be nice.

taiki-e (commented):
Thanks! This looks good to me.

build.rs
```diff
@@ -0,0 +1,18 @@
...
+#[cfg(windows)]
+fn _test2(_: std::os::windows::io::OwnedSocket) {}
+}";
+
+fn main() {
+    let cfg = autocfg::new();
+
+    if cfg.probe_expression(TEST_EXPR) {
+        autocfg::emit("has_io_safety");
+    }
```
taiki-e, line 17: I would prefer to invert cfg here. See [slab's build script](https://github.com/tokio-rs/slab/blob/master/build.rs#L15-L20) for example and reason.

taiki-e: Also, calling rustc in a build script to compile code is relatively fragile (does not work in some host environments and does not work on tier 3 targets). So I would recommend checking the version.

If you need older nightly support, you can consider [this](https://github.com/tokio-rs/tokio/blob/master/tokio/build.rs#L26-L44) approach.

taiki-e (approved):
Thanks!

taiki-e:
Published in 1.8.0.

PR #88, opened 2022-08-21
Add basic benchmarking

notgull:
I'd like to be able to tell if #85 positively or negatively impacts the performance of I/O (although my initial expectations is no effect at all). This PR adds a set of benchmarks that measure the performance of a number of operations.

Currently a work in progress that only measures reading/writing, and ideally would probably only measure one at a time.


notgull:
I have all of the benchmarks I'd like here now.

taiki-e (approved):

PR #91, opened 2022-09-07
Enable support for wasm by disabling io primitives

### LINKED ISSUE #89: Optionally disable timers or Asyncs on features
WASM targets don't support `Async` and probably won't until WASI reaches a more stable point. However, since `parking` is usable on WASM thanks to atomics, it may be desirable to use the `Timer` implementation.

My suggestion is to add two new features: `io` and `timers`. They compose as such:

- With both features enabled (they are by default), the crate functions as normal.
- With `io` disabled but `timers` enabled, the `Reactor` retains its timer processing capabilities but instead of calling to `polling`, it calls `parking` (which works with WASM) with the timeout resulting from `process_timer_ops`.
- With both features disabled, `Reactor` no longer exists and `block_on` is just an alias to the ones in `futures_lite`.

This would be a breaking change.

notgull:
Resolves #89. ~~There are now two features, enabled by default: `io` and `timer`. `io` enables the `Async` type and `timer` enables the `Timer` type. A different reactor is selected depending on which system we're using. When both features are disabled, the implementation of `block_on` is changed to be a brief wrapper around `futures_lite`'s version of it.~~

This is a ~~breaking~~ change.


notgull:
I rebased on master, implemented the changes I made in #92 in this branch, and reimplemented the new system such that it only exposes the fallback when compiled for `target_family = "wasm"`

Cargo.toml
```diff
@@ -19,16 +19,22 @@ name = "io"
...
-socket2 = { version = "0.4.2", features = ["all"] }
 waker-fn = "1.1.0"
 
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
+polling = { version = "2.0.0"}
+slab = { version = "0.4.2" }
+socket2 = { version = "0.4.2", features = ["all"] }
+
+[target.'cfg(target_family = "wasm")'.dependencies]
+wasm-bindgen-test = "0.2"
```
taiki-e, line 36: I think this should be dev-deps.

notgull:
This PR has gotten stale, I'll make a new one for the current state.

PR #95, opened 2022-09-28
Use once_cell from async-lock

### LINKED ISSUE #93: async-io now fails to compile on 1.46
With the new release of [`once_cell` 1.15](https://github.com/matklad/once_cell/pull/198#issuecomment-1249886636), the MSRV had been increased to 1.56 and the edition has been bumped to 2021. This means `async-io`, which depends on `once_cell`, no longer builds on 1.46:

```
$ cargo +1.46 check         
warning: unused manifest key: package.rust-version
    Updating crates.io index
  Downloaded once_cell v1.15.0
error: failed to parse manifest at `/home/jtnunley/.cargo/registry/src/github.com-1ecc6299db9ec823/once_cell-1.15.0/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
```

I see two options:

* Pin our `once_cell` version to `1.14.x`. This would keep our MSRV at 1.46, but prevent us from receiving any new `once_cell` updates.
* Bump our MSRV to 1.56. However, we would no longer be able to build this project on unmodified Debian.

See also: matklad/once_cell#201

notgull:
Resolves #93 by using the `OnceCell` type found in `async-lock`. This type allows for blocking operations, just like the `once_cell::Lazy` type that was used earlier. In addition, `async-lock` does not bump our MSRV, while `once_cell` does.

~~This is a draft PR because smol-rs/async-lock#27 has not yet been released~~

### DIFF
--- a/src/reactor.rs
+++ b/src/reactor.rs
@@ -78,8 +78,6 @@ impl Reactor {
                 events: Mutex::new(Vec::new()),
                 timers: Mutex::new(BTreeMap::new()),
                 timer_ops: ConcurrentQueue::bounded(1000),
-                #[cfg(loom)]
-                mock_events: ConcurrentQueue::unbounded(),
             }
         })
     }
--- a/src/reactor.rs
+++ b/src/reactor.rs
@@ -16,9 +16,9 @@ use std::sync::{Arc, Mutex, MutexGuard};
 use std::task::{Context, Poll, Waker};
 use std::time::{Duration, Instant};
 
+use async_lock::OnceCell;
 use concurrent_queue::ConcurrentQueue;
 use futures_lite::ready;
-use once_cell::sync::Lazy;
 use polling::{Event, Poller};
 use slab::Slab;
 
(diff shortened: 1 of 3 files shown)


notgull:
Clippy failure is unrelated to this pull request.

taiki-e:
Published a new version of async-lock including https://github.com/smol-rs/async-lock/pull/27.

notgull:
> Published a new version of async-lock including [smol-rs/async-lock#27](https://github.com/smol-rs/async-lock/pull/27).

Alright, I've switched to using the released version of `async-lock` in this PR, and marked it ready for review.

src/reactor.rs
```diff
@@ -76,9 +78,10 @@ impl Reactor {
                 events: Mutex::new(Vec::new()),
                 timers: Mutex::new(BTreeMap::new()),
                 timer_ops: ConcurrentQueue::bounded(1000),
+                #[cfg(loom)]
+                mock_events: ConcurrentQueue::unbounded(),
```
taiki-e, line 82: It seems `cfg(loom)` is unrelated.

notgull: Sorry, that was a leftover of an attempt to solve #78

taiki-e (approved):

PR #96, opened 2022-10-25
Fix clippy warnings

notgull:


notgull:
MSRV failure is due to #93

taiki-e (approved):

PR #98, opened 2022-10-29
Use stabilized Duration::MAX for timer

RinHizakura:
It looks like the [`Duration::MAX`](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX) is stabilized in Rust version 1.53.0. Is it possible to change the `duration_max` to `Duration::MAX` now?


notgull:
We would like to avoid bumping the MSRV past 1.48, since that's the current version used by Debian Stable.

RinHizakura:
Thanks, I see. I'll close this now for that consideration.

PR #107, opened 2022-12-30
Reduce syscalls in Async::new

taiki-e:
See https://github.com/smol-rs/async-process/pull/31#discussion_r1059223352 for the context.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ harness = false
 
 [dependencies]
 async-lock = "2.6"
+cfg-if = "1"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -20,6 +20,7 @@ harness = false
 
 [dependencies]
 async-lock = "2.6"
+cfg-if = "1"
```
taiki-e, line 23: (polling already depends on cfg-if, so this doesn't increase number of dependencies)

notgull (approved):

PR #111, opened 2023-01-26
m: Replace socket2 code with rustix

notgull:
This PR allows us to remove our dependency on `socket2` by replacing the equivalent code with functions from `rustix`.

Blocked on bytecodealliance/rustix#495


notgull:
I've just realized that this will add unsafe code on Windows, which I'd like to avoid if at all possible.

PR #112, opened 2023-02-03
feat: Expose other kqueue filters

notgull:
This PR builds on the work done in smol-rs/polling#83 in order to be able to wait on `kqueue` events aside from file descriptor readiness. It accomplishes this by replacing the `raw` field in `Source` with a `Registration` field that, on `kqueue` platforms, accommodates all of the currently defined filters. This allows things like signals and `Child`s to be handled by `Async<T>` as well.

Draft until there is a new release of `polling`. I also need to test it on a BSD virtual machine.


notgull:
The new `polling` version is out now! I've marked this as ready

src/reactor.rs
```diff
@@ -490,7 +480,7 @@ impl<T> Future for Readable<'_, T> {
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         ready!(Pin::new(&mut self.0).poll(cx))?;
-        log::trace!("readable: fd={}", self.0.handle.source.raw);
+        log::trace!("readable: fd={:?}", &self.0.handle.source.registration);
```
fogti, line 483: I think it would be better to give `Registration` an adjusted `Debug` or `Display` implementation instead of using `fd={:?}`, which seems like it would generate a bit confusing output.

fogti: Also, using `&` shouldn't be necessary afaik?

src/reactor/registration.rs
```diff
@@ -0,0 +1,185 @@
+#[cfg(windows)]
```
fogti, line 1: imo it would be a good idea to split this into one file per differing implementation in order to make it easier to compare them side-by-side. (and maybe replace the module hack by imports via paths)

notgull:
Now that I look at it, maybe tying this to `Async<T>` isn't the best idea. The API is kind of obtuse this way, since you need to call `readable()`, and `writable()` will just silently hang forever. Maybe it would be better to have another type that just has a single `ready()` method.

Thoughts? Maybe we can call this type `Filter<T>`.

fogti:
> `writable()` will just hang forever

That sounds pretty bad, it should terminate either with {an I/O error or a panic} if possible...

src/os/kqueue.rs
```diff
@@ -1,58 +1,217 @@
...
+    /// ```
+    /// use async_io::os::kqueue::{Exit, Filter};
+    ///
+    /// # futures_lite::future::block_on(async {
+    /// let child = std::process::Command::new("sleep").arg("5").spawn().unwrap();
+    /// let process = Filter::new(Exit::new(child)).unwrap();
+    /// let inner = process.get_ref();
+    /// # });
+    /// ```
+    pub fn get_ref(&self) -> &T {
```
fogti, line 104: implementing `AsRef` and `AsMut` instead imo might be a better idea.

src/reactor/kqueue.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn from(signal: Signal) -> Self {
+        Self::Signal(signal)
+    }
+}
+
+impl From<Child> for Registration {
+    fn from(process: Child) -> Self {
+        Self::Process(process)
+    }
+}
```
fogti, line 54: is that really necessary?

fogti (changes requested):
imo all one-lines which can end up in a hot path (e.g. `Registration::modify`) should be marked `#[inline]`

src/os/kqueue.rs
```diff
@@ -1,58 +1,217 @@
...
     }
 }
 
 /// Objects that can be registered into the reactor via a [`Async`](crate::Async).
-pub trait Filter: FilterSealed {}
+///
+/// These objects represent other filters associated with the `kqueue` runtime aside from readability
+/// and writability. Rather than waiting on readable/writable, they wait on "readiness". This is
+/// typically used for signals and child process exits.
+pub trait Queueable: QueueableSealed {}
```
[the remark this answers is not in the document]
notgull: Blanket implementing `Queueable` would lead to the docs looking like this:

image

As opposed to the manual implementation, which looks like this:

image

Manual implementation makes the docs more discoverable, IMO.

src/reactor/kqueue.rs
notgull: The first one is necessary, since `RawFd` needs to implement `Into<Registration>`. The others, not so much. Fixed!

fogti (approved):

PR #116, opened 2023-03-17
Change wepoll to IOCP

notgull:
Since smol-rs/polling#88 has been released, we no longer use wepoll.


taiki-e:
Would you update README.md as well?

taiki-e (approved):

PR #119, opened 2023-03-21
v1.12.1

notgull:
- Use [`rustix`] instead of [`libc`]/[`windows-sys`] for system calls (#76)
- Add a `will_fire` method to `Timer` to test if it will ever fire (#106)
- Reduce syscalls in `Async::new` (#107)
- Improve the drop ergonomics of `Readable` and `Writable` (#109)

[`rustix`]: https://crates.io/crates/rustix/
[`libc`]: https://crates.io/crates/libc/
[`windows-sys`]: https://crates.io/crates/windows-sys/


taiki-e:
I think this release should be 1.13.0 because it added a new API.

notgull:
> I think this release should be 1.13.0 because it added a new API (will_fire).

Oh, you're right. Forgot semantic versioning for a second there. Fixed!

PR #122, opened 2023-03-24
chore: use Duration::MAX instead of duration_max

bokuweb:


notgull:
Blocked until Debian Stable starts shipping a newer version of Rust, see smol-rs/smol#224

bokuweb:
i see.thanks.

PR #125, opened 2023-04-06
feat: Add a way to get the global poller's fd

### LINKED ISSUE #124: Expose a way to get the file descriptor of the inner poller
Would solve smol-rs/smol#94

notgull:
Closes #124 by adding a method `async_io::os::unix::reactor_fd()` that allows one to get the file descriptor to register `async-io`'s reactor into an external runtime.


zeenix (approved):
LGTM

PR #128, opened 2023-04-09
fix: typo in comment inside reactor.rs

brahms116:


notgull (changes requested):
I think it was supposed to be "this" instead of "that".

notgull (approved):
Thanks!

PR #135, opened 2023-06-05
Use rustix abstraction to support I/O Safety unconditionally

joshtriplett:
rustix (via io-lifetimes) always provides the I/O Safety types and
traits. On newer Rust it re-exports them from the standard library; on
older Rust it defines its own. We already depend on those types (e.g.
`rustix::fd::BorrowedFd`) in our implementation.

Use the rustix types so that we can support I/O Safety unconditionally.

Note that on Windows, rustix exports `BorrowedFd`/`OwnedFd` types that
are really `BorrowedSocket`/`OwnedSocket`. Work around this in the
implementation by renaming them back.


notgull (commented):
Thanks!

In less than a week from now, Debian Bookworm will release as the new Debian Stable, containing a `rustc` with version 1.63. This means that we can use this version as per our tentative MSRV policy (see smol-rs/smol#244). I would rather just wait and then use `AsFd` and friends from `libstd`, rather than re-export types from `rustix`, which tends to be very unstable.

taiki-e:
Closing in favor of #136

PR #137, opened 2023-06-11
Use Duration::MAX instead of the duration_max hack

notgull:
cc smol-rs/smol#244


notgull:
The build error is fixed in #136

taiki-e (approved):

PR #142, opened 2023-07-11
Update this crate to use the new polling breaking changes

notgull:
For now, this exists as a testing ground in order to test https://github.com/smol-rs/polling/pull/123 before we release it. Consumers of `async-io` should tell me if this API fits their purposes.

Supersedes #136


notgull:
(Sorry for the accidental ping, thought this was another issue)

Cargo.toml
```diff
@@ -54,3 +51,7 @@ timerfd = "1"
 
 [target.'cfg(windows)'.dev-dependencies]
 uds_windows = "1"
+
+[patch.crates-io]
+polling = { git = "https://github.com/smol-rs/polling.git", branch = "notgull/unsafe2" }
+async-io = { path = "." }
```
fogti, line 57: why does this patch `async-io` itself?

notgull: It pulls in `async-net` as a dev dependency, which in turn pulls `async-io`. If the `async-io` version pulled here is the crates.io version it fails because the API of `polling` had a breaking change.

ivmarkov:
@notgull wondering if there is any work still open in this PR that precludes it from merging.

As I understand it, we are in a bit of a weird state right no, and probably we don't want to keep it that way for too long:
* `polling` mainline is already migrated to the new metaphor
* `async-io` still needs this merged, and released versions of `async-io` currently cannot be patched to work against latest `polling` mainline, due to the changes in the `polling` API contract

FWIW, the changes in this branch work fine for me on ESP IDF, but this was not an extensive test in any way.

notgull:
There are a few more breaking changes that I want to make in `polling` first.

examples/linux-inotify.rs
```diff
@@ -37,17 +37,21 @@ fn main() -> std::io::Result<()> {
...
-            .add(".", WatchMask::ALL_EVENTS)?;
+
+        // SAFETY: We do not move the inner file descriptor out.
+        unsafe {
+            inotify
+                .get_mut()
+                .watches()
+                .add(".", WatchMask::ALL_EVENTS)?;
+        }
         println!("Watching for filesystem events in the current directory...");
```
sehz, line 48: this should be using log (tracing?) instead of printing

notgull: Yes, but that's not really relevant to this PR. Feel free to open a new one to fix this.

notgull:
Now that the new `polling` is released, this is ready now.

zeenix:
@notgull cool. CI is failing though.

examples/windows-uds.rs
```diff
@@ -6,54 +6,6 @@
...
-        .await?;
-
-        Ok(())
-    })
-}
-
-#[cfg(not(windows))]
 fn main() {
-    println!("This example works only on Windows!");
+    println!("TODO: Restore this example from git once uds_windows implements I/O safety");
```
zeenix, line 10: I think the PR description and commit log should mention that uds_windows support is being dropped.

zeenix (dismissed):

src/lib.rs
```diff
@@ -1136,14 +1143,45 @@ impl<T> Drop for Async<T> {
...
 
-impl<T: Read> AsyncRead for Async<T> {
+/// Types whose I/O trait implementations do not drop the underlying I/O source.
+///
+/// # Safety
+///
+/// Any I/O trait implementations for this type must not drop the underlying I/O source.
+pub unsafe trait IoSafe {}
+
+// Reference types can't be mutated.
```
fogti, line 1153: what about mutex types and such?

fogti (approved):

src/lib.rs
notgull: Since it's through an `&` I don't think interior mutability can invalidate any existing `AsFd` references. I've added documentation clarifying this.

src/lib.rs
```diff
@@ -1136,14 +1147,73 @@ impl<T> Drop for Async<T> {
...
 }
 
-impl<T: Read> AsyncRead for Async<T> {
+/// Types whose I/O trait implementations do not drop the underlying I/O source.
+///
+/// The resource contained inside of the [`Async`] cannot be invalidated. This invalidation can
+/// happen if the inner resource (the [`TcpStream`], [`UnixListener`] or other `T`) is moved out
+/// and dropped before the [`Async`]. Because of this, functions that grant mutable access to
+/// the inner type are unsafe, as there is no way to guarantee that the source won't be dropped
+/// and a dangling pointer won't be left behind.
```
taiki-e, line 1156: I feel the term "pointer" here is a bit odd because I/O safety is about fd.

src/lib.rs
```diff
@@ -1136,14 +1147,73 @@ impl<T> Drop for Async<T> {
...
+/// system relies on similar guarantees, I believe that this approach is robust.
+///
+/// [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
+/// [`Seek`]: https://doc.rust-lang.org/std/io/trait.Seek.html
+/// [`Write`]: https://doc.rust-lang.org/std/io/trait.Write.html
+///
+/// [`AsyncRead`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncRead.html
+/// [`AsyncSeek`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncSeek.html
+/// [`AsyncWrite`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncWrite.html
+pub unsafe trait IoSafe {}
```
taiki-e, line 1183: I may be saying something silly because I haven't thought too much about it:

- All functions we call within the mut operation are safe and safe functions cannot violate the soundness of other safe code.
- The safety requirement for the `BorrowedFd` returned by `as_fd` is that the fd is not invalid while `BorrowedFd` is alive; if the fd is not valid when `as_fd` is called, it cannot create a `BorrowedFd` to return. (Perhaps they need to panic.)
- If the fd returned before and after the mut operation is different, the original fd may have been replaced. The fd we registered may already be invalid.

Assuming that the fd cannot be concurrently accessed by another thread during the mut operation, couldn't safety here be handled by making sure that the fd returned by `as_fd().as_raw_fd()` matches before and after the mut operation? (I guess we also need a guard to handle that in case the fd is invalidated, though.)

Something like:

```rs
let fd = self.get_ref().as_fd().as_raw_fd();

// mut operation...

let guard = ... // 
assert_eq!(fd, self.get_ref().as_fd().as_raw_fd());
mem::forget(guard);
```

src/lib.rs
fogti: "handle" would make more sense

src/lib.rs
fogti: huh but interior mutability is exactly about being able to invalidate existing inner contents (or at least modifying them concurrently so we can't trust them); I think these "&" impls should require that `T: IoSafe`.

src/lib.rs
notgull: I feel that "pointer" gets across why it's a bad idea to hold the handle after drop in a more serious and direct way. But I'm fine with "handle".

src/lib.rs
notgull: That's a good idea, but unfortunately I don't think it works out. The main issue is what to do if it turns out that the file descriptor is different.

- You can't `delete` the source from the `Poller`. In the time after the original source was dropped, a new source could be created (potentially in another thread) with the same FD and registered into the `Poller`. Doing this deregisters the new source from the `Poller` and results in hard to track down bugs. While it's not *technically* undefined behavior, it's not robust either.
- At the same time, you can't *not* `delete` the source from the `Poller`. Let's say the file descriptor slot hasn't been reallocated yet. Once the `async-io` thread runs `poller.poll()` with a dangling handle it immediately invokes unsound and undefined behavior. Practically this is an issue for Windows and `poll()`-syscall backends, as they both maintain user-space lists of poll-able handles and deleting them like this can lead to random errors or, worse, memory corruption. It has also been observed to be an issue in `kqueue` for some BSDs. Real life effects can include random spurious events and `Waker`s never being woken.

Essentially we're forced into a [Catch-22](https://en.m.wikipedia.org/wiki/Catch-22_(logic)) where, if we're unsure of the current fate of the file descriptor, the only available options lead to buggy and unsound code. Therefore we have to make it an invariant of the `Async` type that the file descriptor is alive at all times.

src/lib.rs
notgull: The thing is that it's impossible to return a file descriptor-wielding resource from within any current interior mutability container in Rust. This code doesn't work for example:

```rust
struct Foo(RefCell<TcpStream>);

impl AsFd for Foo {
    fn as_fd(&self) -> BorrowedFd<'_> {
        self.0.borrow().as_fd()
        // ^ borrow checker throws an error on this line
    }
}
```

This is because the borrow checker  recognizes that the contents of `foo.0` can be changed during the lifetime of the `BorrowedFd` and thus prevents it from passing. The only primitive where this kind of thing would work is `OnceCell`, and the main reason why it works is because `OnceCell` can't be mutated after the value is first instantiated.

Pinning guarantees operate on generally the same principle.

The worst thing that can happen is that external state is used to change what kind of pointer `as_fd()` returns. For instance:

```rust
struct Bar {
    flag: Cell<bool>,
    a: TcpStream,
    b: TcpStream
}

impl AsFd for Bar {
    fn as_fd(&self) -> BorrowedFd<'_> {
        if self.flag.replace(!self.flag.get()) {
            &self.a
        } else {
            &self.b
        }
    }
}
```

We solve this problem by only calling `as_fd()` once to get the original source. Implementations like this are considered buggy (but not unsound) and are thus not really supported by `async-io`.

src/lib.rs
fogti: well, then panic! if that happens (because we can assume that we can't recover the `Poller` easily)

notgull: We can't `panic!` because other threads (and even the current thread) can keep going after a panic, which would result in even more undefined behavior. The proper thing to do in this case would be to `abort()`. However I would like to avoid aborting if possible, as that brings down the entire program which is disastrous for long running programs like web servers. I think we would be better citizens of the Rust ecosystem if we just make this precondition unsafe rather than trying to paper over the unsafety with panics or aborts.

fogti: well, we could `abort()` when build in debug mode...

notgull: I guess we could...

src/lib.rs
```diff
@@ -1136,14 +1147,74 @@ impl<T> Drop for Async<T> {
...
+/// happen if the inner resource (the [`TcpStream`], [`UnixListener`] or other `T`) is moved out
+/// and dropped before the [`Async`]. Because of this, functions that grant mutable access to
+/// the inner type are unsafe, as there is no way to guarantee that the source won't be dropped
+/// and a dangling handle won't be left behind.
+///
+/// Unfortunately this extends to implementations of [`Read`] and [`Write`]. Since methods on those
+/// traits take `&mut`, there is no guarantee that the implementor of those traits won't move the
+/// source out while the method is being run.
+///
+/// This trait is an antidote to this predicament. By implementing this trait, it is guaranteed
```
fogti, line 1162: wouldn't "pledged" be a better wording than "guaranteed"? Or perhaps "By implementing this trait, the implementer certifies that using any I/O traits won't destroy the source"....

btw. s/desroy/destroy/ .

src/lib.rs
fogti: "The worst thing that can happen" is exactly what I meant. Ok. (imo the last paragraph of your response should be added as a comment into the source)

src/lib.rs
taiki-e: @notgull 

Thanks for the explanation. IIUC, it appears that the bad cases you mention here is that it can be reproduced even if you wrap the underlying Io object with a Mutex or something.

taiki-e: Nah, if using Mutex or something, the problem may not really occur with safe code because of the guard's lifetime issue.

src/lib.rs
```diff
@@ -1136,14 +1147,74 @@ impl<T> Drop for Async<T> {
...
+///
+/// [`AsyncRead`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncRead.html
+/// [`AsyncSeek`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncSeek.html
+/// [`AsyncWrite`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncWrite.html
+pub unsafe trait IoSafe {}
+
+// Reference types can't be mutated.
+unsafe impl<T: ?Sized> IoSafe for &T {}
+unsafe impl<T: ?Sized> IoSafe for std::rc::Rc<T> {}
+unsafe impl<T: ?Sized> IoSafe for Arc<T> {}
```
taiki-e, line 1188: Wouldn't this implementation make it possible to bypass safety requirements using methods like get_mut?

notgull: In theory you can't implement your own `Read`/`Write`/`Seek` on these types so it shouldn't be an issue. But a future version of Rust could undo that guarantee, so I'll remove it just to be safe.

src/lib.rs
notgull: > well, we could `abort()` when built in debug mode...

This would require some code restructuring that I think should be reserved for a future PR. I'll resolve this convo for now.

src/lib.rs
```diff
@@ -1136,14 +1147,96 @@ impl<T> Drop for Async<T> {
...
+/// [`AsyncSeek`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncSeek.html
+/// [`AsyncWrite`]: https://docs.rs/futures-io/latest/futures_io/trait.AsyncWrite.html
+pub unsafe trait IoSafe {}
+
+/// Reference types can't be mutated.
+///
+/// The worst thing that can happen is that external state is used to change what kind of pointer
+/// `as_fd()` returns. For instance:
+///
+/// ```no_compile
```
taiki-e, line 1190: Did you mean `compile_fail`?

src/lib.rs
```diff
@@ -1136,14 +1147,96 @@ impl<T> Drop for Async<T> {
...
+/// source out while the method is being run.
+///
+/// This trait is an antidote to this predicament. By implementing this trait, the user pledges
+/// that using any I/O traits won't destroy the source. This way, [`Async`] can implement the
+/// `async` version of these I/O traits, like [`AsyncRead`], [`AsyncWrite`] and [`AsyncSeek`].
+///
+/// # Safety
+///
+/// Any I/O trait implementations for this type must not drop the underlying I/O source. Traits
+/// affected by this trait include [`Read`], [`Write`] and [`Seek`].
```
taiki-e, line 1169: Is there a reason `BufRead` is not listed here?

taiki-e (commented):
LGTM

src/lib.rs
notgull: `Async<T>` does not currently implement `AsyncBufRead` if `T` is `BufRead`, so I didn't include it. Although I would be fine with adding it if you think we'd support it in the future.

src/lib.rs
notgull: This involves `BorrowedFd` and I didn't want to make this test conditional on Windows.

src/lib.rs
taiki-e: Given that we mention Seek/AsyncSeek (which we do not currently implement) here, I think it is reasonable to mention BufRead/AsyncBufRead as well.

src/lib.rs
notgull: Actually, nevermind, I figured out a fix for this.

PR #147, opened 2023-09-07
Update rustix version

rodoufu:
# Version 1.12.1

- Update `rustix`

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -57,6 +57,5 @@ openbsd_task:
   setup_script:
     # OpenBSD is tier 3 target, so install rust from package manager instead of rustup
     - pkg_add git rust
-    - rustup update
   test_script:
     - cargo test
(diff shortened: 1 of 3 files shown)


rodoufu:
openbsd is failing with:
```sh
package `clap_lex v0.5.1` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.68.0
Either upgrade to rustc 1.70.0 or newer, or use
cargo update -p clap_lex@0.5.1 --precise ver
where `ver` is the latest version of `clap_lex` supporting rustc 1.68.0
```
Since it is installed with `pkg_add` and not `rustup` I'm not quite sure how to fix that.

notgull (commented):
The OpenBSD build error is fixed in #142.

Is there anything noteworthy in Rustix v0.38.11 that justifies upgrading to it? If not I'd rather avoid merging this, since Cargo will automatically select later versions of Rustix in the same breaking release given the option.

taiki-e:
I would prefer not to raise the minimal versions unless they fail to compile with the current minimal versions of the dependencies.

[cargo's "x.y.z" requirement means "^x.y.z" (caret) requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html), and the user can just run `cargo update` to get the latest semver-compatible version.

So, I'll close this, but thanks for the PR anyway.

PR #148, opened 2023-09-09
refactor: remove unnecessary `Option` in `Async`

cjwcommuny:
In this pull request, I push down the `drop` method of `Async` to its `source` field. This allows for the replacement of the `Option<T>` with `T`, which is not possible in the current implementation due to the restriction [E0509](https://doc.rust-lang.org/error_codes/E0509.html).

This refactoring efficiently eliminates several instances of `unwrap`. Moreover, it reduces the repetition of the `Reactor::get().remove_io(&self.0)`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,6 +70,7 @@ use std::sync::Arc;
 use std::task::{Context, Poll, Waker};
 use std::time::{Duration, Instant};
 
+use std::ops::Deref;
 #[cfg(unix)]
 use std::{
     os::unix::io::{AsFd, AsRawFd, BorrowedFd, OwnedFd, RawFd},
@@ -619,10 +620,10 @@ impl Stream for Timer {
 #[derive(Debug)]
 pub struct Async<T> {
     /// A source registered in the reactor.
-    source: Arc<Source>,
+    source: ArcSource,
 
     /// The inner I/O handle.
-    io: Option<T>,
+    io: T,
 }
 
 impl<T> Unpin for Async<T> {}
@@ -679,8 +680,8 @@ impl<T: AsFd> Async<T> {
         let registration = unsafe { Registration::new(fd) };
 
         Ok(Async {
-            source: Reactor::get().insert_io(registration)?,
-            io: Some(io),
+            source: ArcSource(Reactor::get().insert_io(registration)?),
+            io,
         })
     }
 }
@@ -760,8 +761,8 @@ impl<T: AsSocket> Async<T> {
         let registration = unsafe { Registration::new(borrowed) };
 
         Ok(Async {
-            source: Reactor::get().insert_io(registration)?,
-            io: Some(io),
+            source: ArcSource(Reactor::get().insert_io(registration)?),
+            io: io,
         })
     }
 }
@@ -813,7 +814,7 @@ impl<T> Async<T> {
     /// # std::io::Result::Ok(()) });
     /// ```
     pub fn get_ref(&self) -> &T {
-        self.io.as_ref().unwrap()
+        &self.io
     }
 
     /// Gets a mutable reference to the inner I/O handle.
@@ -834,7 +835,7 @@ impl<T> Async<T> {
     /// # std::io::Result::Ok(()) });
     /// ```
     pub unsafe fn get_mut(&mut self) -> &mut T {
-        self.io.as_mut().unwrap()
+        &mut self.io
     }
 
     /// Unwraps the inner I/O handle.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1965,3 +1953,21 @@ fn connect(addr: SockAddr, domain: Domain, protocol: Option<Protocol>) -> io::Re
...
+    }
+}
+
+impl Deref for ArcSource {
+    type Target = Source;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
```
notgull, line 1973: I'd rather have this in `reactor.rs` and named something a little better, like `SourceContainer` or something.

notgull (changes requested):
Looks good to me aside from one minor detail

src/lib.rs
cjwcommuny: done

notgull (commented):
Thanks! This looks good to me now.

Before you merge, do you mind adding a test for this functionality? It would be inline in `reactor.rs`. Just test to make sure that, when an `Async<T>` is registered and then dropped, the key in no longer present in the sources list.

cjwcommuny:
> Thanks! This looks good to me now.
> 
> Before you merge, do you mind adding a test for this functionality? It would be inline in `reactor.rs`. Just test to make sure that, when an `Async<T>` is registered and then dropped, the key in no longer present in the sources list.

Done.

notgull (approved):
Thanks! The tests are remarkably well constructed.

Once you rebase on `master` I'll merge this.

cjwcommuny:
> Thanks! The tests are remarkably well constructed.
> 
> Once you rebase on `master` I'll merge this.

Done.

src/lib.rs
```diff
@@ -679,8 +679,8 @@ impl<T: AsFd> Async<T> {
         let registration = unsafe { Registration::new(fd) };
 
         Ok(Async {
-            source: Reactor::get().insert_io(registration)?,
-            io: Some(io),
+            source: SourceContainer(Reactor::get().insert_io(registration)?),
```
notgull, line 682: Rather than doing this, I'd prefer for `SourceContainer` to implement `From<Arc<Source>>`. That way it can be constructed with `.into()` instead of manually like this.

src/lib.rs
```diff
@@ -855,9 +855,8 @@ impl<T> Async<T> {
     /// inner.set_nonblocking(false)?;
     /// # std::io::Result::Ok(()) });
     /// ```
-    pub fn into_inner(mut self) -> io::Result<T> {
-        let io = self.io.take().unwrap();
-        Reactor::get().remove_io(&self.source)?;
+    pub fn into_inner(self) -> io::Result<T> {
+        let Self { source: _, io } = self;
```
notgull, line 859: I just realized, this no longer returns an error if `remove_io()` fails. Can you fix it so that the error isn't ignored here?

notgull (changes requested):
Noticed a few more things on a second read-through.

src/lib.rs
cjwcommuny: Done.

src/reactor.rs
```diff
@@ -673,3 +674,47 @@ impl<H: Borrow<crate::Async<T>>, T> Drop for Ready<H, T> {
...
+mod test {
+    use crate::reactor::Reactor;
+    use crate::Async;
+
+    use std::net::TcpListener;
+
+    #[test]
+    fn test_async_drop() {
+        let async_io =
+            Async::<TcpListener>::bind(([127, 0, 0, 1], 0)).expect("failed to create async io");
```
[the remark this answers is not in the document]
cjwcommuny: Done.

src/lib.rs
cjwcommuny: Done.

src/lib.rs
```diff
@@ -855,9 +855,10 @@ impl<T> Async<T> {
     /// inner.set_nonblocking(false)?;
     /// # std::io::Result::Ok(()) });
     /// ```
-    pub fn into_inner(mut self) -> io::Result<T> {
-        let io = self.io.take().unwrap();
-        Reactor::get().remove_io(&self.source)?;
+    pub fn into_inner(self) -> io::Result<T> {
+        let Self { mut source, io } = self;
+        source.remove()?;
+        std::mem::forget(source);
```
notgull, line 861: This leaks the `Arc` and everything inside. I think you have to either:

1). Make `SourceContainer` a wrapper around `Option<Arc<Source>>` and have `remove()` take it out.
2). Make `Async` just contain `Option<SourceContainer>`.

notgull (changes requested):

src/lib.rs
cjwcommuny: It appears that satisfying these two prerequisites proves to be an impasse: 

1. Bypassing the usage of `Option`
2. Preserving the semantics of the method `Async::into_inner`

Therefore, the only logical step forward is to close this pull request. I appreciate all your reviews.

PR #153, opened 2023-09-23
Web timer support

### LINKED ISSUE #89: Optionally disable timers or Asyncs on features

notgull:
This commit adds support for async-io on wasm32-unknown-unknown. Not all
features of async-io can be ported to WASM; for instance:

- Async<T> can't be ported over as WASM doesn't really have a reactor. WASI
  could eventually be supported here, but that is dependent on
  https://github.com/smol-rs/polling/issues/102
- block_on() can't be ported over, as blocking isn't allowed on the web.

The only thing left is Timer, which can be implemented using setTimeout and
setInterval. So that's what's been done: when the WASM target family is enabled,
Async<T> and block_on() will be disabled and Timer will switch to an
implementation that uses web timeouts.

This is not a breaking change, as this crate previously failed to compile on web
platforms anyways.

This functionality currently does not support Node.js.

Closes #89


src/lib.rs
```diff
@@ -61,38 +61,34 @@
...
 
+#[cfg(not(target_family = "wasm"))]
 mod driver;
+#[cfg(not(target_family = "wasm"))]
+mod io;
+#[cfg(not(target_family = "wasm"))]
 mod reactor;
 
+#[cfg_attr(not(target_family = "wasm"), path = "timer/native.rs")]
+#[cfg_attr(target_family = "wasm", path = "timer/web.rs")]
```
taiki-e, line 82: This is actually only necessary in `target_os = "unknown"` cases, right?

notgull: WASI doesn't have support in `polling` yet... although WASI also doesn't really have support for web primitives either, so compiling to it should really be a compile error.

As for `emscripten`, I don't know whether it supports `poll` or `epoll` or anything, so until we figure that out we should be using web timers.

So you're right, this should be `any(target_os = "unknown", target_os = "emscripten")`.

notgull: ....actually I'm pretty sure that `not(target_os = "wasi")` covers our bases better.

taiki-e:
https://github.com/smol-rs/async-io/actions/runs/6358096463/job/17270021140#step:11:44

```
     Running unittests src/lib.rs (target/wasm32-unknown-unknown/debug/deps/async_io-1a608e2b038c59[41](https://github.com/smol-rs/async-io/actions/runs/6358096463/job/17270021140#step:11:42).wasm)
no tests to run!
     Running tests/async.rs (target/wasm32-unknown-unknown/debug/deps/async-6cc94a735e74c985.wasm)
                                                  
no tests to run!
     Running tests/block_on.rs (target/wasm32-unknown-unknown/debug/deps/block_on-0[44](https://github.com/smol-rs/async-io/actions/runs/6358096463/job/17270021140#step:11:45)b18783949e5c2.wasm)
                                                  
no tests to run!
     Running tests/timer.rs (target/wasm32-unknown-unknown/debug/deps/timer-2[47](https://github.com/smol-rs/async-io/actions/runs/6358096463/job/17270021140#step:11:48)dda665b5d544f.wasm)
                                                  
this test suite is only configured to run in a browser, but we're only testing node.js tests so skipping
```

taiki-e:
> This functionality currently does not support Node.js.

It probably does not work in Web Worker either. https://github.com/rustwasm/wasm-bindgen/issues/1046

Also, there may be an issue mentioned here. https://github.com/tomaka/wasm-timer/issues/21#issuecomment-1484553707
(It would make sense to mention in the documentation if such a problem exists.)

daxpedda:
The [Prioritized Task Scheduling API](https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API) is currently the solution to this problem, but is unfortunately only supported by Chrome so far, but Firefox [is working on it](https://mozilla.github.io/standards-positions/#scheduling-api).

The nested call throttling of `setTimeout()` and `setInterval()` can be circumvented by using a `MessagePort` (I didn't find anything better so far) to queue either.

See the Winit implementation as an example:
- [`setTimeout()`](https://github.com/rust-windowing/winit/blob/ee0db52ac49d64b46c500ef31d7f5f5107ce871a/src/platform_impl/web/web_sys/schedule.rs#L105-L142)
- [`postTask()`](https://github.com/rust-windowing/winit/blob/ee0db52ac49d64b46c500ef31d7f5f5107ce871a/src/platform_impl/web/web_sys/schedule.rs#L59-L88)

> It probably does not work in Web Worker either.

This can be circumvented by providing your own bindings, e.g. I've done this in [`web-time`](https://github.com/daxpedda/web-time/blob/940b9c5357374f20701e2037080e5482aa03ca19/src/web/js.rs).

notgull:
I would prefer having wider browser support. I'll add a note to the documentation for `Timer` clarifying that on web platforms it can take longer.

daxpedda:
> I would prefer having wider browser support.

You can do runtime support detection, we have done this in [Winit](https://github.com/rust-windowing/winit/blob/f5dd1c008ceb13309d98eb1333b56bc92f973ace/src/platform_impl/web/web_sys/schedule.rs#L163-L183).

VitalyAnkh:
How is this PR going?

PR #149, opened 2023-09-09
Optimize block_on by caching Parker and Waker

Pixelstormer:
`async-io`'s implementation of `block_on` currently creates a new parker and waker on every invocation. This PR caches the parker and waker in a thread-local to improve performance. Most of this code is copied from [`futures-lite`'s implementation of `block_on`](https://github.com/smol-rs/futures-lite/blob/2234456b71cd5d20d67c3ee4c5e3c97a25912612/src/future.rs#L56), as that crate already does exactly this, and the performance difference between these two `block_on` implementations is why I looked into optimizing `async-io` in the first place.

<details><summary>futures-lite (Red) vs async-io without this PR (Yellow)</summary>
<p>

futures-lite vs async-io

</p>
</details> 

<details><summary>futures-lite (Red) vs async-io with this PR (Yellow)</summary>
<p>

futures-lite vs patched async-io

</p>
</details> 

(Also see https://github.com/bevyengine/bevy/pull/9626 for more context)

PR #140, opened 2023-06-14
Replace log with tracing

notgull:
cc smol-rs/smol#244, smol-rs/polling#119, smol-rs/blocking#40

Replaces `log` with `tracing`

PR #155, opened 2023-09-25
v2.0.0

notgull:
- **Breaking:** `Async::new()` now takes types that implement `AsFd`/`AsSocket` instead of `AsRawFd`/`AsRawSocket`, in order to implement I/O safety. (#142)
- **Breaking:** `Async::get_mut()`, `Async::read_with_mut()` and `Async::write_with_mut()` are now `unsafe`. The underlying source is technically "borrowed" by the polling instance, so moving it out would be unsound. (#142)
- Expose miscellaneous `kqueue` filters in the `os::kqueue` module. (#112)
- Expose a way to get the underlying `Poller`'s file descriptor on Unix. (#125)
- Add a new `Async::new_nonblocking` method to allow users to avoid duplicating an already nonblocking socket. (#159)
- Remove the unused `fastrand` and `memchr` dependencies. (#131)
- Use `tracing` instead of `log`. (#140)
- Support ESP-IDF. (#144)
- Optimize the `block_on` function to reduce allocation, leading to a slight performance improvement. (#149)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# Version 2.0.0
+
+- **Breaking:** `Async::new()` now takes types that implement `AsFd`/`AsSocket` instead of `AsRawFd`/`AsRawSocket`, in order to implement I/O safety. (#142)
+- **Breaking:** `Async::get_mut()`, `Async::read_with_mut()` and `Async::write_with_mut()` are now `unsafe`. The underlying source is technically "borrowed" by the polling instance, so moving it out would be unsound. (#142)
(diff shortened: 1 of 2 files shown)


zeenix:
> * **Breaking:** `Async::get_mut()`, `Async::read_with_mut()` and `Async::write_with_mut()` are now `unsafe`. The underlying source is technically "borrowed" by the polling instance, so moving it out would be unsound. (https://github.com/smol-rs/async-io/pull/143)

I think there is a mistake here, #143 has nothing to do with the description here.

notgull:
> I think there is a mistake here, #143 has nothing to do with the description here.

Fixed!

Issue #132, opened 2023-05-28
Support creating `Async` without setting non-blocking mode


h33p:
I have a project aims to be pluggable into different async executors through FDs/Handles. It works great, I have a PoC working with `tokio`. It also works on linux with `smol`. [Example of how I'm using `Async`](https://github.com/memflow/mfio/blob/main/mfio/src/backend/integrations/async_io.rs#L78).

However, on macOS I'm getting `ENOTTY`, which stems from `Async::new` unconditionally running `fcntl` to set O_NONBLOCK, even on FDs that do not support it (like kqueue).

My usecase of `Async` is purely to have async-io poll for read-readiness of the FD, and have my task woken up. That's the whole extent of it. Maybe it would make sense to expose a different type just for polling, but the simpler way would be to add an overload for `Async::new` that allows making `fcntl` call conditional.

PR #159, opened 2023-10-08
Add Async::new_nonblocking and avoid needless set_nonblocking calls

taiki-e:
connect() already sets socket to non-blocking mode.

The second comment exposes Async::new_nonblocking to resolve https://github.com/smol-rs/async-io/issues/132.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -636,8 +636,8 @@ impl<T: AsFd> Async<T> {
     /// This method will put the handle in non-blocking mode and register it in
     /// [epoll]/[kqueue]/[event ports]/[IOCP].
     ///
-    /// On Unix systems, the handle must implement `AsRawFd`, while on Windows it must implement
-    /// `AsRawSocket`.
+    /// On Unix systems, the handle must implement `AsFd`, while on Windows it must implement
+    /// `AsSocket`.
     ///
     /// [epoll]: https://en.wikipedia.org/wiki/Epoll
     /// [kqueue]: https://en.wikipedia.org/wiki/Kqueue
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -731,16 +733,18 @@ impl<T: AsSocket> Async<T> {
...
-        let borrowed = io.as_socket();
-
         // Put the socket in non-blocking mode.
-        set_nonblocking(borrowed)?;
+        set_nonblocking(io.as_socket())?;
+
+        Self::new_nonblocking(io)
+    }
 
+    fn new_nonblocking(io: T) -> io::Result<Async<T>> {
```
notgull, line 742: Can we make this public to solve https://github.com/smol-rs/async-io/issues/132?

notgull (approved):
Thanks!

src/lib.rs
taiki-e: Done in https://github.com/smol-rs/async-io/pull/159/commits/de07b6adca6432e2fde5ce2a15c66c41731d7992.

src/lib.rs
```diff
@@ -731,16 +750,35 @@ impl<T: AsSocket> Async<T> {
...
+
+        Self::new_nonblocking(io)
+    }
 
+    /// Creates an async I/O handle without setting it to non-blocking mode.
+    ///
+    /// This method will register the handle in [epoll]/[kqueue]/[event ports]/[IOCP].
+    ///
+    /// On Unix systems, the handle must implement `AsRawFd`, while on Windows it must implement
+    /// `AsRawSocket`.
```
notgull, line 764: Nit: This is supposed to be `AsFd` and `AsSocket`.

taiki-e: Good catch. Fixed.

notgull (approved):
Thanks!

Issue #154, opened 2023-09-25
Make sure we can't deregister non-registered sources


notgull:
Calloop recently found a bug where it's possible to deregister a source by trying to register it twice. See https://github.com/Smithay/calloop/pull/153 for more information.

I don't *think* `async-io` is affected by this bug, but we should add a test case to make sure.

notgull:
Closed in #161

PR #161, opened 2023-10-08
Add test for #154

siqpush:
Guard against duplicate source registrations

### DIFF
--- a/tests/async.rs
+++ b/tests/async.rs
@@ -1,8 +1,11 @@
+use std::cell::RefCell;
 use std::future::Future;
 use std::io;
 use std::net::{Shutdown, TcpListener, TcpStream, UdpSocket};
+use std::os::fd::AsFd;
 #[cfg(unix)]
 use std::os::unix::net::{UnixDatagram, UnixListener, UnixStream};
+use std::rc::Rc;
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


notgull:
You need to fix the Windows tests by using `&socket1` instead of `socket1.borrow().as_fd()`.

tests/async.rs
```diff
@@ -356,3 +359,41 @@ fn shutdown() -> io::Result<()> {
         Ok(())
     })
 }
+
+// prevent source from unregistering by trying to register it twice
+#[test]
+fn duplicate_socket_insert() -> io::Result<()> {
+    future::block_on(async {
+        let (socket1, socket2) = match Async::<UnixStream>::pair() {
```
notgull, line 367: Instead of this, use the `tcp_pair` function in this module.

tests/async.rs
```diff
@@ -356,3 +356,33 @@ fn shutdown() -> io::Result<()> {
...
+// prevent source from unregistering by trying to register it twice
+#[test]
+fn duplicate_socket_insert() -> io::Result<()> {
+    future::block_on(async {
+        let listener = Async::<TcpListener>::bind(([127, 0, 0, 1], 0))?;
+        let addr = listener.as_ref().local_addr()?;
+
+        // attempt to register twice
+        assert!(
+            Async::<TcpListener>::bind(addr).is_err(),
```
notgull, line 369:
```suggestion
            Async::<TcpListener>::new(&listener).is_err(),
```

siqpush: @notgull only additional change to suggestion is removing the type annotation

notgull (approved):
Thanks!

Issue #165, opened 2023-10-29
Remove the waker-fn dependency


notgull:
It should be replaced by an object that implements `Wake`.

PR #171, opened 2023-11-05
Remove dependency on waker-fn

asonix:
Resolves https://github.com/smol-rs/async-io/issues/165

Adds a new concrete type `BlockOnWaker` that implements `Wake`

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -10,7 +10,6 @@ use std::time::{Duration, Instant};
 use async_lock::OnceCell;
 use futures_lite::pin;
 use parking::Parker;
-use waker_fn::waker_fn;
 
 use crate::reactor::Reactor;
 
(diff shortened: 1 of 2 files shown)


src/driver.rs
```diff
@@ -131,19 +130,38 @@ pub fn block_on<T>(future: impl Future<Output = T>) -> T {
...
-                    }
+        let waker = BlockOnWaker::create(io_blocked.clone(), u);
+
+        (p, waker, io_blocked)
+    }
+
+    struct BlockOnWaker {
+        io_blocked: Arc<AtomicBool>,
+        unparker: parking::Unparker,
+    }
```
notgull, line 141: Could you move the structure and it's implementation after the `thread_local` declaration? Since it uses the thread local I think it will be less confusing for readers if it comes after.

notgull (commented):
Looks good to me aside from one minor nitpick.

notgull (approved):
Thanks!

PR #176, opened 2023-12-06
Add support for abstract sockets

notgull:
In #146, I introduced a bug that prevented abstract sockets from
working. I passed the path straight into
rustix::net::SocketAddrUnix::new, which fails if it receives an abstract
socket.

This commit fixes this issue by explicitly checking for abstract
sockets. If it sees that the path it's receiving is abstract, it will
pass the path's bytes to new_abstract_socket() instead.

This should fix the issue that is occurring in dbus2/zbus#517

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1847,24 +1847,7 @@ impl Async<UnixStream> {
     /// # std::io::Result::Ok(()) });
     /// ```
     pub async fn connect<P: AsRef<Path>>(path: P) -> io::Result<Async<UnixStream>> {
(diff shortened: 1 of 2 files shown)


zeenix (approved):
LGTM. Thanks so much for looking into this. Would be great to have a test case for abstract sockets btw.

notgull:
> LGTM. Thanks so much for looking into this. Would be great to have a test case for abstract sockets btw.

Added a test, and in the process found a bug that I'd missed.

zeenix (approved):
:+1:

PR #180, opened 2024-01-07
Use const thread_local & Fix CI failure

taiki-e:
- Use const thread_local
  This is available since Rust 1.59 (https://github.com/rust-lang/rust/pull/91355).
- Fix CI failure due to FreeBSD 12.4 EoL
- Use cargo-hack's --rust-version flag for msrv check
  This respects rust-version field in Cargo.toml, so it removes the need to manage MSRV in both the CI file and Cargo.toml.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,7 +14,7 @@ env:
 freebsd_task:
   name: test ($TARGET)
   freebsd_instance:
-    image_family: freebsd-12-4
+    image_family: freebsd-13-2
   matrix:
     - env:
         TARGET: x86_64-unknown-freebsd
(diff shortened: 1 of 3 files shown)


taiki-e:
```
pkg install -y git
Updating FreeBSD repository catalogue...
pkg: http://pkgmir.geo.freebsd.org/FreeBSD:12:amd64/quarterly/meta.txz: Not Found
repository FreeBSD has no meta file, using default settings
pkg: http://pkgmir.geo.freebsd.org/FreeBSD:12:amd64/quarterly/packagesite.pkg: Not Found
pkg: http://pkgmir.geo.freebsd.org/FreeBSD:12:amd64/quarterly/packagesite.txz: Not Found
Unable to update repository FreeBSD
Error updating repositories!
```

Oh, FreeBSD 12 was EoL in 2023-12...

PR #187, opened 2024-03-05
Fix SocketFlags usage on AIX

ecnelises:
AIX has neither SOCK_NONBLOCK nor SOCK_CLOEXEC flags.


taiki-e (approved):
Thanks!

notgull:
@ecnelises Do you know of a way to test AIX locally? I don't have AIX so I can't run it in a VM to verify this.

PR #196, opened 2024-08-05
Set nosigpipe on NetBSD and DragonFly BSD

taiki-e:
Follow-up to https://github.com/smol-rs/async-io/pull/146#discussion_r1318111984.

rustix 0.38.18+ supports it (https://github.com/bytecodealliance/rustix/pull/870) and std now sets it on these targets (https://github.com/rust-lang/rust/pull/124470#issuecomment-2081489885).


notgull (approved):
Thanks!

PR #202, opened 2024-09-14
Add support for visionOS

madsmtm:
[In the future](https://github.com/rust-lang/rust/issues/100343#issuecomment-2065159934) you'll probably be able to use `target_family = "darwin"` instead, but for now `target_vendor = "apple"` should suffice.

Similar to https://github.com/smol-rs/polling/pull/217.


taiki-e (approved):
Thanks!

PR #203, opened 2024-09-16
feat: remove async-lock in favor of OnceLock

rogercoll:
Is there any reason why using `async-lock` and not the standard library thread safe OnceCell (OnceLock)?

My guess was that the `async-lock` dependency was used because the OnceLock struct was not stable yet, [since 1.70.0](https://doc.rust-lang.org/std/sync/struct.OnceLock.html)

Rust version bumped to 1.70.0, what is the strategy to follow on Rust updates?


notgull:
Our MSRV is currently pinned to Debian Stable, which is at 1.63.0:

rogercoll:
Thanks! I will close the PR up until then.

PR #206, opened 2024-10-06
Bump slab to v0.4.7

notgull:
cc smol-rs/async-executor#131


taiki-e:
AFAIK, async-io doesn't use slab APIs that requires 0.4.7, but it is okay with using the same version requirements to keep things simple.

taiki-e:
Closing per https://github.com/smol-rs/async-io/pull/206#issuecomment-2395519295 (now confirmed by the check added in https://github.com/smol-rs/async-io/commit/c86a47783435ed2c5adfe970d353c32996a56e3a).

PR #207, opened 2024-10-12
refactor: remove child ownership from Exit filter

rogercoll:
These change loose the BSD systems API when creating an Exit struct. At the moment, the Exit struct takes ownership of the Child struct and not publicly exporting it, preventing being able to call Child's methods.  

The Exit struct is primarily used to wait for a child process termination using the reactor's registration functionality. The Registration uses the polling crate to perform the async wait. The polling only requires a PID, thus these changes removes the child ownership from the package in favor of just providing a PID number.

This will ease the development of the `async-process` for BSD systems: https://github.com/smol-rs/async-process/issues/49#issuecomment-2385345337

### DIFF
--- a/examples/kqueue-process.rs
+++ b/examples/kqueue-process.rs
@@ -14,36 +14,24 @@
     target_os = "dragonfly",
 ))]
 fn main() -> std::io::Result<()> {
-    use std::{num::NonZeroI32, process::Command};
+    use std::process::Command;
 
     use async_io::os::kqueue::{Exit, Filter};
     use futures_lite::future;
 
     future::block_on(async {
         // Spawn a process.
-        let mut process = Command::new("sleep")
+        let process = Command::new("sleep")
             .arg("3")
             .spawn()
             .expect("failed to spawn process");
 
         // Wrap the process in an `Async` object that waits for it to exit.
-        let process_handle = unsafe {
-            Filter::new(Exit::from_pid(
-                NonZeroI32::new(process.id().try_into().expect("invalid process pid"))
-                    .expect("non zero pid"),
-            ))?
-        };
+        let process = Filter::new(Exit::new(process))?;
 
         // Wait for the process to exit.
-        process_handle.ready().await?;
+        process.ready().await?;
 
-        println!(
-            "Process exit code {:?}",
-            process
-                .try_wait()
-                .expect("error while waiting process")
-                .expect("process did not exit yet")
-        );
         Ok(())
     })
 }
--- a/examples/kqueue-process.rs
+++ b/examples/kqueue-process.rs
@@ -14,24 +14,34 @@
     target_os = "dragonfly",
 ))]
 fn main() -> std::io::Result<()> {
-    use std::process::Command;
+    use std::{num::NonZeroI32, process::Command};
 
     use async_io::os::kqueue::{Exit, Filter};
     use futures_lite::future;
 
     future::block_on(async {
         // Spawn a process.
-        let process = Command::new("sleep")
+        let mut process = Command::new("sleep")
             .arg("3")
             .spawn()
             .expect("failed to spawn process");
 
         // Wrap the process in an `Async` object that waits for it to exit.
-        let process = Filter::new(Exit::new(process))?;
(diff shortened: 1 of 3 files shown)


rogercoll:
Any clue why the CI is failing?
```
   /home/runner/.cargo/bin/cargo audit --json --file ./Cargo.lock
  error: not found: Couldn't load ./Cargo.lock
  Caused by:
    -> I/O operation failed: I/O operation failed: entity not found
```

Does it require the Cargo.lock file from now on?

src/os/kqueue.rs
```diff
@@ -233,23 +239,23 @@ impl QueueableSealed for Signal {
...
 /// it will return a [`readable`](crate::Async::readable) event when the child process exits.
 #[derive(Debug)]
-pub struct Exit(Option<Child>);
+pub struct Exit(NonZeroI32);
 
 impl Exit {
     /// Create a new `Exit` object.
-    pub fn new(child: Child) -> Self {
-        Self(Some(child))
+    pub fn new(pid: NonZeroI32) -> Self {
```
notgull, line 251: I'd like to avoid a breaking change, can we make this a new method instead?

notgull (changes requested):

src/os/kqueue.rs
rogercoll: Good point, a new `from_pid` constructor added in https://github.com/smol-rs/async-io/pull/207/commits/430846412d260388f5375f4e2aec94aa907444c6 instead. I made it `unsafe` to propagate the safety statement when creating a polling process from a pid.

Wdyt?

notgull: I don't think it actually needs to be unsafe? While it's unsafe for FDs, PIDs are allowed to be racy IIRC

rogercoll: The main reason behind the `unsafe` was to propagate the "safety" statement [defined in the polling process](https://github.com/smol-rs/polling/blob/master/src/os/kqueue.rs#L219) (and used by the reactor):

```rust
/// Create a `Process` from a PID.
    ///
    /// # Safety
    ///
    /// The PID must be tied to an actual child process.
    pub unsafe fn from_pid(pid: std::num::NonZeroI32, ops: ProcessOps) -> Self {
        Self {
            pid: unsafe { rustix::process::Pid::from_raw_unchecked(pid.get()) },
            ops,
            _lt: PhantomData,
        }
    }
```

With the new Exit status API, the struct can be crated for any given u32 value, thus the possibility of not being tied to an actual child process (differently than child.id()). This is the error when the provided pid is not attached to a process:
```rust
Error: Os { code: 3, kind: Uncategorized, message: "No such process" }
```

rogercoll: Given that with the new `from_pid` it lets the user provide any PID value, do you still think the unsafe guidance is not needed?

Thanks for your feedback.

examples/kqueue-process.rs
```diff
@@ -14,24 +14,36 @@
...
             .expect("failed to spawn process");
 
         // Wrap the process in an `Async` object that waits for it to exit.
-        let process = Filter::new(Exit::new(process))?;
+        let process_handle = unsafe {
+            Filter::new(Exit::from_pid(
+                NonZeroI32::new(process.id().try_into().expect("invalid process pid"))
+                    .expect("non zero pid"),
+            ))?
+        };
```
notgull, line 35: I would prefer keeping the example simple.

notgull (changes requested):
LGTM aside from one minor nitpick

examples/kqueue-process.rs
rogercoll: sgtm https://github.com/smol-rs/async-io/pull/207/commits/610773a62c852404514c797884696bc1f4201c91

notgull:
I'll try to push this as a release this weekend

rogercoll:
@notgull Is there any ETA for the next release? Thanks!

notgull:
Sorry, I got distracted. Assuming nothing major comes up tonight I'll push a release.

notgull:
...I said, blissfully unaware that something major was about to come up. 😵‍💫

I'll release this ASAP

PR #222, opened 2025-04-28
Temporarily disable openBSD in testing

dcampbell24:


taiki-e:
Thanks for the PR. I'm fine with temporarily disabling OpenBSD test, but I don't want to remove it forever.


The current OpenBSD CI failure is due to the fact that OpenBSD 7.5 used in the image used by CI is now EoL, which should be fixed by updating it. However, since the image is not under our control, temporarily disabling it is a reasonable solution here.

taiki-e (approved):
Thanks!

PR #224, opened 2025-05-08
Include cargo.lock.

dcampbell24:
See https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control for the reasons why.


taiki-e:
The current CI error is upstream issue and will be fixed in next nightly (https://github.com/rust-lang/rust/pull/140734).

PR #225, opened 2025-05-12
v2.4.1

dcampbell24:
Make a release with updated rustix dependency, so that you can build this library on Redox.


dcampbell24:
Someone with permission will have to do the actual release.

taiki-e (approved):
Thanks! I will release this together with the rustix update on `polling` crate which is a dependency of async-io.

PR #227, opened 2025-05-20
Update criterion to 0.6, Update rust-version to 1.66, fix warnings.

dcampbell24:


dcampbell24:
In order to use `black_box` from the `std` you have to update the `rust-version`.

dcampbell24:
Okay, should we leave this open in case some other feature warrants bumping the MSRV?

PR #230, opened 2025-06-13
Update to windows-sys 0.60

zeenix:
Closes #231

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,11 @@ defaults:
 jobs:
   fmt:
     uses: smol-rs/.github/.github/workflows/fmt.yml@main
+  clippy:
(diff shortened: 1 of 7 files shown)


zeenix:
Hmm.. seems it requires bumping MSRV to 1.65. @taiki-e what's our MSRV policy again?

taiki-e:
>   error: package `tracing-core v0.1.34` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is 1.63.0

I think `tracing` dependency should be optional.
That is what we did in the past in `blocking` (https://github.com/smol-rs/blocking/pull/60) and what we are trying to do now in `polling` (https://github.com/smol-rs/polling/issues/231).

taiki-e (approved):

PR #237, opened 2025-08-04
impl IoSafe for std::io::PipeReader & std::io::PipeWriter

jk-ozlabs:
It would be handy to have async wrappers for the pipe types (from std::io::pipe()), so impl IoSafe for both.

The pipe types were introduced in 1.87, but the MSRV for async-io appears to be 1.63 currently. So, protect the impls with a cfg(has_pipe), and use autocfg to set has_pipe, like was done in a20076f ("Implement I/O-safe traits (#84)")

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,16 @@
+fn main() {
+    let cfg = match autocfg::AutoCfg::new() {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            println!("cargo:warning=async-io: failed to detect compiler features: {e}");
+            return;
+        }
+    };
+
+    // We use "no_*" instead of "has_*" here. For (non-Cargo) build processes
+    // that don't run build.rs, the negated version gives us a recent
+    // feature-set by default.
+    if !cfg.probe_rustc_version(1, 87) {
+        autocfg::emit("async_io_no_pipe");
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ categories = ["asynchronous", "network-programming", "os"]
 exclude = ["/.*"]
 
 [lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(polling_test_poll_backend)'] }
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(polling_test_poll_backend)', 'cfg(async_io_no_pipe)' ] }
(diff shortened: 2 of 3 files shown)


jk-ozlabs:
Looks like `emit_possibility` was only in `autocfg` 1.5, so I have split into a simpler `emit_has_type()` check, that has earlier compat with the minimum versions test.

This means we have two `cfg()`s, but they directly align with the type availability.

build.rs
```diff
@@ -0,0 +1,17 @@
...
+            return;
+        }
+    };
+
+    // emit expected config settings; may be replaced with autocfg::emit_possibility
+    // if we update to autocfg 1.5
+    println!("cargo:rustc-check-cfg=cfg(has_std__io__PipeWriter)");
+    println!("cargo:rustc-check-cfg=cfg(has_std__io__PipeReader)");
+
+    cfg.emit_has_type("std::io::PipeReader");
```
taiki-e, line 15: Could you use a build script in the same format as https://github.com/smol-rs/polling/pull/39's one? Specifically:

- Calling rustc in a build script to compile code (which emit_has_type does) is relatively fragile (does not work in some host environments and does not work for tier 3 targets). So I would recommend checking the version.
- I would prefer to invert cfg because [it is also works well with non-cargo build systems](https://github.com/smol-rs/polling/pull/39#issuecomment-1237278819).

taiki-e (changes requested):
Thanks!

jk-ozlabs:
> Thanks!

and thanks for the review! :smile:

build.rs
```diff
@@ -0,0 +1,20 @@
...
+        Ok(cfg) => cfg,
+        Err(e) => {
+            println!("cargo:warning=async-io: failed to detect compiler features: {e}");
+            return;
+        }
+    };
+
+    // emit expected config settings; may be replaced with
+    // autocfg::emit_possibility if we update to autocfg>=1.5
+    println!("cargo:rustc-check-cfg=cfg(async_io_no_pipe)");
```
notgull, line 12: This can be moved to `Cargo.toml`, right?

jk-ozlabs: Yep, I am happy to implement it there if preferred.

taiki-e (approved):

taiki-e:
Published in 2.6.0.

PR #239, opened 2025-08-05
Add Timer::pause

### LINKED ISSUE #238: stop timer
right now the only way to stop a timer seems to be setting a period that is so long so adding it to `Instant::now()` overflows. I think `Duration::MAX` should do it but I don't know if it's guaranteed. can we also provide a method to clearly pause a timer without any room for doubt?

yshui:
Closes #238

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -342,7 +342,6 @@ impl Timer {
             None => {
                 // Overflow to never going off.
                 self.clear();
-                self.when = None;
             }
         }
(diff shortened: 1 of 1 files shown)


yshui:
@notgull updated. i don't know if setting `self.when = None;` is necessary, but i did that to be consistent with the other cases of turing-timer-offs.

notgull (approved):
Thanks!

PR #240, opened 2025-08-10
Bump MSRV to 1.70 & Drop async_lock & Replace futures_lite::{ready,pin,future::poll_fn}

taiki-e:
It is likely that MSRV will rise to 1.85 in the near future due to rustix, but for now, let's raise it to the version that is actually necessary.

cc https://github.com/smol-rs/smol/issues/244


notgull (approved):
Thanks!

PR #242, opened 2025-08-25
Add doc aliases `sleep` and `timeout` to `Timer`

janriemer:
Two examples are shown in [`Timer`'s docs](https://docs.rs/smol/latest/smol/struct.Timer.html):
One example for basic _sleeping_ and one example for having a _timeout_ on a future.

This adds corresponding doc aliases to `Timer`, so that this struct is easier to find with those use cases in mind.

See doc alias feature:
https://blog.rust-lang.org/2020/11/19/Rust-1.48/#adding-search-aliases


taiki-e:
CI issue is recently fixed https://github.com/rust-lang/rust/issues/145952.

PR #249, opened 2026-02-15
feat: Avoid spawning async-io thread if possible

### LINKED ISSUE #40: Avoid spawning the async-io thread if possible

notgull:
This commit adds conditions that prevent the async-io thread from being
spawned in certain cases. This is important because sometimes async-io
is used in environments where spawning additional threads can interfere
with other functionality in the program; for example, in DPDK poll
loops.

If there is only one async-io::block_on invocation, there is no need to
spawn the "async-io" thread; we can just modify block_on to not give up
the reactor if it can detect that it is the only block_on call running
in the program. From there, we also need to avoid spawning the async-io
thread if there are no resources left when block_on exits. This should
allow async-io to not spawn additional threads in single-threaded
programs where it is not needed.

Currently draft while I work out the kinks.

Closes: #40

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,8 +45,10 @@ autocfg = "1"
 [dev-dependencies]
 async-channel = "2.0.0"
 async-net = "2.0.0"
+async-task = "4.7.1"
 blocking = "1"
 criterion = { version = "0.7", default-features = false, features = ["cargo_bench_support"] }
+fastrand = "2.3.0"
 getrandom = "0.3"
 signal-hook = "0.3"
 tempfile = "3"
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -1,7 +1,7 @@
 use std::cell::{Cell, RefCell};
 use std::future::Future;
 use std::pin::pin;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::atomic::{fence, AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, OnceLock};
 use std::task::{Context, Poll, Waker};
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -45,8 +45,10 @@ autocfg = "1"
 [dev-dependencies]
 async-channel = "2.0.0"
 async-net = "2.0.0"
+async-task = "4.7.1"
 blocking = "1"
 criterion = { version = "0.7", default-features = false, features = ["cargo_bench_support"] }
+fastrand = "2.3.0"
```
fogti:
```suggestion
fastrand = "2.3"
```

fogti (commented):
It looks quite good overall (I couldn't find any obvious mistakes), but I'm not familiar enough to vouch for its complete correctness.

src/driver.rs
```diff
@@ -118,12 +148,20 @@ pub fn block_on<T>(future: impl Future<Output = T>) -> T {
...
+        // if it isn't already.
+        let _ = unparker();
+    }
 
     // Make sure to decrement `BLOCK_ON_COUNT` at the end and wake the "async-io" thread.
     let _guard = CallOnDrop(|| {
         BLOCK_ON_COUNT.fetch_sub(1, Ordering::SeqCst);
-        unparker().unpark();
+
+        // Only wake the `async-io` thread iff there are live resources.
```
zeenix: typo: iff.

zeenix (changes requested):

